PR #97, opened 2026-01-21
PR39: branch1 memtable profiling kickoff

snissn:
PR39: branch1 kickoff (memtable path profiling)

Summary
- Adds mode3 vlogprof harness to capture steady-state CPU profile for dict-on writes.
- Updates compression_copy_optimization runbook with append-only work log and initial branch1 profile notes.

Tests
- go test ./cmd/unified_bench -count=1
- go test -tags vlogprof ./cmd/unified_bench -run TestProfileVlogDict_Mode3_DictOn_Ultra_1024 -count=1 -v
- go test -tags vlogprof ./cmd/unified_bench -run TestProfileVlogDict_Mode4_DictOn_Ultra_1024 -count=1 -v

### DIFF
--- a/slab-optimization/compression_copy_optimization.md
+++ b/slab-optimization/compression_copy_optimization.md
@@ -200,6 +200,72 @@ mode3 (journal on) and mode4 (journal off).
 Reduce copy amplification and allocations inside the value‑log writer, independent of memtable vs
 streaming path, for both dict‑on and dict‑off modes.
 
+### Autopilot Runbook (agent-friendly)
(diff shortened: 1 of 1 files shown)


snissn:
Update: reverted the dict direct-append optimization (commit 627efaa1b6) because it caused valuelog corruption.
New regression test TestDictAppendReadRoundTrip (TreeDB/internal/valuelog) now passes and will catch future regressions.
Current diff vs PR38 is mostly docs + bench harness + tests.

PR #231, opened 2026-01-30
Add cached bloat regression test (bench phases + vacuum shrink)

snissn:
This targets runaway `index.db` growth + slow checkpoints seen in `unified-bench` with `-treedb-keep-recent 1`.

Key changes
- `TreeDB/freelist`: add chunked `Allocator.FreeMany` + tests (amortizes freelist head page updates without long mutex holds).
- `TreeDB/db/prune.go`: switch prune batches to `FreeMany`.
- `TreeDB/db/batch.go`: run a small bounded prune pass (`currentSeq=baseSeq+1`) before `Apply` when `KeepRecent<=1` so eligible pages can be reused *before* allocator starvation forces append growth.
- `TreeDB/lifecycle/graveyard.go`: allow reclaim at the min pinned seq boundary (`seq == minPinned` eligible) + regression test.
- Add overwrite-churn regression test in cached mode.
- Update freelist reuse tests to reflect KeepRecent eligibility timing.

Local perf smoke
Command:

```
make unified-bench && time ./bin/unified-bench \
  -test batch_write,random_write,batch_delete \
  -dbs treedb -profile fast -keys 2000000 -format markdown \
  -checkpoint-between-tests -treedb-keep-recent 1
```

Results (macOS):
- Batch Write: ~2.84M ops/s
- Random Write: ~72.8k ops/s
- Batch Delete: ~3.46M ops/s

Checkpoint time between tests:
- before Random Write: ~3.72s
- before Batch Delete: ~2.32s

Tests
- `go test ./...`

Notes
- `TreeDB/BLOAT_FIX_PLAN.md` documents the “middle path” approach (prune/reuse first, bounded locking, then revisit higher-level structure only if needed).

### DIFF
--- a/TreeDB/caching/db.go
+++ b/TreeDB/caching/db.go
@@ -1614,7 +1614,7 @@ type DB struct {
 	testOnVlogSync  func(laneID int)
 }
 
-func (db *DB) flushBackendEntriesCap(totalOps int) int {
+func (db *DB) flushBackendEntriesCap(totalOps int, sync bool) int {
 	capEntries := db.flushBackendMaxEntries
 	if capEntries < 1 {
 		capEntries = 1
@@ -1622,6 +1622,12 @@ func (db *DB) flushBackendEntriesCap(totalOps int) int {
 	maxBatches := db.flushBackendMaxBatches
 	if maxBatches == 0 {
 		maxBatches = 16
+	}
+	// Sync-triggered flushes (checkpoint/close) should remain fast; cap the number
+	// of intermediate commits in that path even if the steady-state micro-batch
+	// policy is aggressive.
+	if sync && maxBatches > 8 {
+		maxBatches = 8
 	}
 	if maxBatches > 0 && totalOps > capEntries*maxBatches {
 		// Increase the chunk size so we emit at most maxBatches intermediate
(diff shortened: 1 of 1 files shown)


snissn:
Implemented a concrete coalesce/merge fix:

- Change: `TreeDB/zipper/zipper.go` `coalesceLeafChildren`
  - Empty-leaf pruning (first pass) now runs even when the maintenance budget is exhausted.
  - If budget is exhausted, we *skip* the more expensive sibling merge/rebalance pass, but keep the pruning result.

Effect (validated by tests):
- `TestCachedDeleteAll_CollapsesToSmallTree` flips from red -> green.
- `TestCachedBenchBloat_PhaseFragmentationTimeline` now shows `after_batch_delete` collapses to ~6 user pages instead of ~1560 near-empty leaves.

Remaining bloat signal:
- The live-tree pages after (re)write are still ~2.3x a vacuum rebuild, which now looks like a fill-factor / incremental-build structural issue rather than empty-leaf retention.

snissn:
More evidence on the "underfilled pages" mechanism:

- Added passing diagnostic `TestCachedSeed_BatchSizeAffectsFillAndPageCount` (`TreeDB/caching/seed_batchsize_fill_test.go`).
  - Seeding 20k keys via many small batches yields ~1594 live pages with leaf fill avg ~0.43.
  - Seeding the same 20k keys in a single big batch yields ~694 live pages with leaf fill avg ~0.99 (vacuum-like).
  - This shows the bloat is strongly correlated with *commit batch size*, not just deletes.

- Added passing experiment `TestCachedBenchBloat_PageCountRatioVsVacuum_WithForcedMaintenance` (`TreeDB/caching/freelist_bloat_forced_maintenance_test.go`).
  - Even when we force maintenance=true for put-only batches and disable maintenance budgeting (`MaintenanceOpsPerCoalesce=0`), the pagecount ratio vs vacuum remains ~2.29.
  - So simply "turning on maintenance" does not compact the small-batch structure; likely we need either:
    - an append/bulk-load path for monotonic ingest, or
    - an explicit compaction pass (vacuum-lite) on checkpoint/close, or
    - different split / separator-key behavior so later batches keep filling the last partially-full leaf.

snissn:
Added a new failing regression to force us to fix the small-batch underfilled-page behavior:

- `TestCachedSeed_BatchSizeFillShouldBeHigh` (`TreeDB/caching/seed_batchsize_fill_threshold_test.go`)
  - Sweeps batch sizes (default 256,512,1024,4096,20000)
  - Asserts `treedb.user.leaf_fill_ppm_avg >= 800000` for each
  - Current failures (expected today): 256/512/1024/4096 are ~0.43 avg fill; 20000 is ~0.99.

This should make the "compact like vacuum regardless of commit size" requirement explicit and guard against regressions once we implement a fix (bulk-load/append-path, improved split packing, or compaction on checkpoint/close).

snissn:
PR #231: cached bloat regression test + related reuse/prune tweaks — partially addresses #251 (regression coverage), but is CONFLICTING.

Issue #267, opened 2026-02-04
Sprint 2: Template encode CPU/alloc optimization (routing + mask fast path)


snissn:
## Goal
Make template encoding materially faster (CPU + allocations) for the workloads where it *does* match, focusing on routing/fingerprinting and mask-template fast paths.

We’ll keep evaluating throughput vs space savings as we go.

## Context
After Sprint 1 makes template mode safe, Sprint 2 targets the main CPU costs when templates are active.

## Must do
- [ ] **Replace expensive routing fingerprints with cheaper routing for common cases**
  - In `TreeDB/template/fingerprint.go` / routing code used by `Engine.Encode`, avoid full winnowing + map/sort per value when possible.
  - Consider a small fixed set of chunk hashes + length hash (no allocations) as a fast routing key.

- [ ] **Mask-template fast path (reduce per-byte work)**
  - Optimize mask-template matching/encoding to avoid scanning all bytes for every value.
  - Use precomputed var positions and constant spans; verify constants with `memcmp`-style compares and extract var bytes efficiently.

- [ ] **Reduce encode payload allocations**
  - Add pooled/scratch buffers for template payload construction (`EncodePayload` / mask payload construction).
  - Avoid allocating new slices for every encoded value when possible.

## Should do
- [ ] **Make profiling easy and documented**
  - Ensure `TreeDB/cmd/vlog_dict_realdata` supports capturing CPU profiles for the relevant region.
  - Document a short “how to profile template encode” snippet (commands + what to look for) in an appropriate doc or the benchmark README.

## Acceptance criteria
- On a template-shaped dataset:
  - `TemplateOnly` throughput improves materially compared to the pre-sprint baseline.
  - CPU profiles show reduced time in fingerprinting/routing and mask matching.
  - Alloc/op (or total allocs) reduced in the hot path.

## How to measure
- `TreeDB/cmd/vlog_dict_realdata -bench-kv` (template-only) + `-cpu-profile`.
- Compare `steady_raw_MBps` + pprof top stacks before/after.

snissn:
Completed in #274 (merged).

PR #274, opened 2026-02-04
template: Sprint 2 encode CPU/alloc optimizations

snissn:
Implements #267.

## Summary
This targets template-encode CPU + allocation costs when templates are active, with a focus on routing/fingerprinting and mask-template fast paths.

## Changes
- **Routing fingerprints (fast path):** add `AppendRoutingFingerprints` / internal `appendRoutingFingerprints` using a small set of fixed segment hashes (+ length) instead of per-value winnowing. `Engine.Encode` uses a stack buffer (no alloc) and only falls back to legacy winnowed routing when the fast path yields zero candidates.
- **Mask template matching/encoding:** precompute var/const spans at template-def decode time and use `bytes.Equal` over constant spans (memcmp-style) + var-only loops when constants match.
- **Anchor template candidate loop:** avoid allocating gaps/payload for each candidate; use `matchTemplateLen` for scoring and only build the payload for the winning kept candidate.
- **Training overhead:** move `TrainSampleStride` gating before any fingerprint work; make `BucketKey` allocation-free.
- **Profiling UX:** add `-bench-cpu-profile` to `TreeDB/cmd/vlog_dict_realdata` (steady-phase-only), and document template profiling in `docs/TREEDB_BENCHMARKING.md`.

## How to measure
- Template smoke table (4 modes): `scripts/bench_template_smoke.sh`
- Template-only profiling:
  - `go run ./TreeDB/cmd/vlog_dict_realdata -bench-kv -bench-template on -bench-compression off ... -bench-cpu-profile /tmp/treedb_template_steady.pprof`
  - `go tool pprof -http=:0 /tmp/treedb_template_steady.pprof`

Notes:
- Defaults remain unchanged; template compression remains opt-in.

### DIFF
--- a/TreeDB/cmd/vlog_dict_realdata/main.go
+++ b/TreeDB/cmd/vlog_dict_realdata/main.go
@@ -74,6 +74,7 @@ type benchConfig struct {
 	DictBytes         int
 	DictSampleStride  int
 	DictWaitSeconds   int
+	CPUProfileSteady  string
 	OutJSON           string
 	KeepDir           bool
 }
@@ -161,6 +162,7 @@ func main() {
 	benchTemplate := flag.String("bench-template", "off", "Bench template compression: on|off|prepass")
 	benchKeepDir := flag.Bool("bench-keep-dir", false, "Keep bench directory after run")
 	cpuProfile := flag.String("cpu-profile", "", "Write CPU profile to this file (optional)")
+	benchCPUProfile := flag.String("bench-cpu-profile", "", "Write CPU profile for steady-state phase (bench only; optional)")
 	benchRawMiB := flag.Int("bench-raw-mib", 512, "Raw MiB to write in steady-state phase")
 	benchBatch := flag.Int("bench-batch", 1024, "Number of ops per batch write")
 	benchKeyMode := flag.String("bench-key-mode", "random", "Key mode: random|sequential|dataset")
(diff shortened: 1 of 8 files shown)


snissn:
Local perf/profiling notes (template-only, compression off).

Dataset: `scripts/bench_template_smoke.sh` generated JSONL (`/tmp/treedb_template_smoke.jsonl`, deterministic).

Command (both main + this PR):
- `go run ./TreeDB/cmd/vlog_dict_realdata -input /tmp/treedb_template_smoke.jsonl -input-encoding string -train 10000 -eval 5000 -cap 0 -bench-kv -bench-mode wal_off -bench-pointer-threshold 1 -bench-compression off -bench-template on -bench-raw-mib 4096 -bench-batch 1024 -bench-key-mode dataset -cpu-profile /tmp/template.pprof -bench-out-json /tmp/template.json`

Headline steady throughput (single run; likely within noise):
- main: `steady_raw_MBps=519.16`
- PR:   `steady_raw_MBps=522.23`

pprof (single run; top stacks dominated by mask matching/encode as expected). With this PR I no longer see `template.NormalizeConfig` showing up in the top-20 on this workload.

New steady-only profiler flag:
- `-bench-cpu-profile /tmp/treedb_template_steady.pprof` (bench only; avoids profiling load/warmup).

PR #326, opened 2026-02-05
TreeDB/node: shrink columnar leaf headers (implicit offsets)

snissn:
Removes the redundant per-entry `keyOff`/`valOff` fields from columnar leaf entry headers (both columnar-only and columnar+prefix-v2).

Why:
- The encoder already writes value bytes first, then key bytes, so offsets are derivable from header size + value size.
- Saves 4B/entry, improving leaf density and reducing `index.db` size in columnar modes (especially P1 C1 V1).

Notes:
- Pre-alpha storage format: this is an on-disk format change for columnar leaves.

Tests:
- Adds density regression tests to ensure columnar and columnar+prefix+packed do not reduce keys/page for pointer entries.

Bench:
- Will post updated unified-bench numbers (server, median-3-of-5) in a follow-up comment.

### DIFF
--- a/TreeDB/node/builder.go
+++ b/TreeDB/node/builder.go
@@ -584,6 +584,22 @@ func (b *Builder) AddInternalChild(key []byte, childPageID uint64) error {
 	return nil
 }
 
+// AddInternalChildParts appends an internal entry from split key parts.
+// This preserves call sites that already handle prefix/suffix key views.
+func (b *Builder) AddInternalChildParts(prefix, suffix []byte, childPageID uint64) error {
+	if len(prefix) == 0 {
+		return b.AddInternalChild(suffix, childPageID)
+	}
+	keyLen := len(prefix) + len(suffix)
+	if keyLen > int(^uint16(0)) {
+		return ErrKeyTooLarge
+	}
+	key := make([]byte, keyLen)
+	copy(key, prefix)
+	copy(key[len(prefix):], suffix)
+	return b.AddInternalChild(key, childPageID)
+}
+
 // Finish finalizes the page header and checksum.
 // Returns a Node wrapper for convenience.
 func (b *Builder) Finish() *Node {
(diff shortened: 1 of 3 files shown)


snissn:
Server bench (Ubuntu 22.04, i5-11400F, `taskset -c 0-11`), `keys=4,000,000`, `profile=fast`, `-treedb-force-value-pointers`, 5 runs each; reported **median-3-of-5** (drop min/max, avg middle 3).

### P1 C1 V1 (prefix+columnar+packed) + internal base-delta
- Before (PR301 @ `aa89f69`): BW=1,886,865  RW=1,454,905  BD=1,570,110  `maindb/index.db`=192.0 MiB
- After  (this PR @ `11e2894`): BW=1,876,002  RW=1,463,332  BD=1,555,410  `maindb/index.db`=170.7 MiB (**-11.1%**)

### Control: P1 C0 V1 (prefix+packed) + internal base-delta
- This PR @ `11e2894`: BW=1,859,382  RW=1,462,742  BD=1,571,441  `maindb/index.db`=170.7 MiB

So the “why is P1C0V1 smaller than P1C1V1?” mystery is essentially the per-entry 4B offset overhead: after making offsets implicit, **P1C1V1 disk usage matches P1C0V1** in this workload.

### P0 C1 V0 (columnar-only) + internal base-delta
- Before (PR301 @ `aa89f69`): BW=1,872,452  RW=1,430,027  BD=1,591,928  `maindb/index.db`=298.7 MiB
- After  (this PR @ `11e2894`): BW=1,865,680  RW=1,459,918  BD=1,557,478  `maindb/index.db`=256.0 MiB

Local sanity check: `BenchmarkLeafPageDensity` now shows `ptr/columnar` == `ptr/plain` and `ptr/columnar_prefix/packed` == `ptr/prefix/packed` in keys/page.

Issue #328, opened 2026-02-05
TreeDB vlog dict: frame planner (target raw bytes per frame) to avoid tiny frames


snissn:
## Context
The current dict frame grouping is driven primarily by a stored/learned `K`. For small values (e.g. 128B) this can produce very small frames (1–4KiB raw), where per-frame overhead (framing, CRC, scheduling, pipeline plumbing) dominates and prevents parallel compression from scaling.

This issue introduces an explicit *frame planner* that can choose frame boundaries based on a **target raw bytes per frame** (i.e. throughput/ratio tuning knob) while still respecting `MaxRecordSize` and `MaxFrameK`.

## Proposal
- Add a new option (and `cmd/unified_bench` flag) to control grouping by target size:
  - `Options.ValueLog.DictFrameTargetBytes` (0 = use existing behavior / dict K)
- When dict compression is enabled and `DictFrameTargetBytes > 0`:
  - Compute `avgValueBytes` over the batch (or use a cheap running estimate).
  - Choose `k = clamp(1..MaxFrameK, DictFrameTargetBytes / max(1, avgValueBytes))`.
  - Clamp by `MaxRecordSize` just like today (`maxKBySize`).
  - Use the computed `k` for frame grouping for that append.
- Keep the on-disk format unchanged.

## Acceptance criteria
- For `valsize=128` and `DictFrameTargetBytes=16KiB`, frames should typically use `k≈128` (subject to clamps).
- No correctness regressions: pointers resolve, reopen tests pass.
- Unit tests cover planner behavior (clamps, max size, varying value sizes).

## Bench / repro
Use `cmd/unified_bench` `batch_write` and `random_read` to validate throughput and read amplification tradeoffs. Example (adjust keys as needed):

```bash
go run ./cmd/unified_bench \
  -dbs treedb \
  -test batch_write,random_read \
  -profile fast \
  -keys 1000000 \
  -valsize 128 \
  -batchsize 100 \
  -progress=false \
  -format markdown \
  -treedb-force-value-pointers \
  -treedb-vlog-dict on \
  -treedb-vlog-dict-frame-target-bytes 16384 \
  -val-pattern ultra_compressible_repeat
```

## Links
- Builds on: #312 (parallel pipeline)

PR #332, opened 2026-02-05
TreeDB vlog dict: frame target bytes planner

snissn:
Implements #328.

Adds `ValueLog.DictFrameTargetBytes` (+ unified_bench flag `-treedb-vlog-dict-frame-target-bytes`) to target an approximate raw payload size per dict-compressed frame. This avoids producing tiny frames for small values where per-frame overhead dominates.

Notes:
- Default behavior unchanged when the option is 0.
- Includes unit tests for the planner + MaxRecordSize clamp.


snissn:
Coordination update: this PR is now part of the RC integration stack on `rc/vlog-dict-throughput` (not targeted for immediate main merge). Exit goal is issue #346: dict-on throughput must beat dict-off on highly compressible 5M gates, with follow-up tracks #347 #348 #349 #350.

Issue #351, opened 2026-02-05
META: RC handoff for TreeDB vlog dict throughput (branches, gates, and remaining concerns)


snissn:
## Objective (north star)
For highly compressible values, TreeDB with value-log dictionary compression enabled should achieve **higher end-to-end write throughput** than the same run with dictionary compression disabled, while preserving strong space savings.

This meta issue tracks all remaining work and provides a complete handoff package for a new agent.

## Current status snapshot
- `main` has already merged the low-risk foundations (`#314-#322`, `#336`, `#343`).
- The active pipeline stack is intentionally **RC-only**, not targeted to `main` yet.
- `#323` (pipeline v1) was closed as superseded.

## Branches and PR topology (current)
- RC integration branch: `rc/vlog-dict-throughput`
- PR stack (in order):
  1. `#332` TreeDB vlog dict frame planner (base: `rc/vlog-dict-throughput`)
  2. `#333` prepared grouped record API (base: `#332` branch)
  3. `#334` pipeline v2 (base: `#333` branch)
  4. `#335` lane vlog queue dict batching (base: `#334` branch)
  5. `#324` docs on top of stack

### Core RC objective
- #346 RC throughput objective gate (dict-on must beat dict-off)

### Throughput and adaptation tracks
- #347 autotune v2 for `TrainBytes` / `DictBytes`
- #348 retrain cadence + data-shift triggers
- #350 dict-on throughput hotspot elimination sprint

### UX/docs track
- #349 one-flag/default dict compression UX and docs clarity

### Existing implementation tracks in this stack
- #328 frame target bytes planner
- #329 prepared grouped record API
- #330 pipeline v2
- #331 lane queue dict batching
- #310 autotune docs/stats context
- #281 degenerate stream activation bug (constant/all-zero stream)

## Validation / performance gates (required)
Use **paired trials** on the same host/session (interleave control and candidate).

### Gate command (5M required)
```bash
go run ./cmd/unified_bench \
  -dbs treedb \
  -test batch_write,prefix_scan \
  -keys 5000000 \
  -valsize 128 \
  -batchsize 100 \
  -progress=false \
  -format markdown \
  -treedb-force-value-pointers \
  -treedb-vlog-dict both \
  -treedb-vlog-compression-autotune medium \
  -val-pattern ultra_compressible_repeat
```

### Trial policy
- Minimum 5 trials.
- Report:
  - median-of-5
  - middle-3 trimmed mean
  - raw per-trial table

### Pass criteria
[...]

PR #362, opened 2026-02-06
TreeDB: reduce vlog dict-on gap for issue #351

snissn:
## Summary
- keep Iteration B checkpoint behavior and tune force-pointer dict training candidate K selection to avoid low-K default bias
- remap default autotune candidate K set from `[1,2,4,8,16,32]` to `[8,16,32]` in force-pointer mode
- add tiny-frame prepared-path no-copy override (`k>=8`, ultra-low entropy) to reduce encode/reset overhead in `FramePreparer`

## Validation
- `go test ./TreeDB/caching ./TreeDB/internal/valuelog -count=1`
- `go test ./TreeDB/... -count=1`
- 5x interleaved timed gate on clean host (`keys=5M`, `valsize=128`, `batchsize=10000`, dict off/on)

## Perf (latest gate)
- commit: `1ae2b1e`
- off median: `1,653,935`
- on median: `1,627,196`
- dict on-vs-off gap: `-1.62%`

Related: #351

### DIFF
--- a/TreeDB/README.md
+++ b/TreeDB/README.md
@@ -94,9 +94,9 @@ db, err := treedb.Open(opts)
 Profiles are intended to make intent explicit:
 
 - `ProfileDurable`: safest defaults (recommended).
-- `ProfileFast`: relax durability/integrity knobs for throughput.
-- `ProfileWALOnFast`: fast ingest profile that keeps WAL on but relaxes durability checks.
-- `ProfileBench`: deterministic benchmarking profile (not production).
+- `ProfileFast`: relax durability/integrity knobs for throughput and enables index optimizations (`ForcePointers`, `LeafPrefixCompression`, `IndexColumnarLeaves`, `IndexPackedValuePtr`, `IndexInternalBaseDelta`).
+- `ProfileWALOnFast`: fast ingest profile that keeps WAL on, relaxes durability checks, and enables the same index optimizations.
+- `ProfileBench`: deterministic benchmarking profile (not production); includes `ProfileFast` index optimizations.
 
 Note: WAL-off is selected via `opts.Durability = treedb.DurabilityWALOffRelaxed`.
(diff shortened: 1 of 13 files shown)


snissn (commented):
Decision: approve with revisions.

Findings (ordered by severity):

1) Potential shutdown hang in dict prep pipeline
- `prepareAppendDictFrames` waits for an exact `frameCount` results using blocking receives (`TreeDB/caching/db.go:4465`), while worker goroutines may exit as soon as `closeCh` is closed (`TreeDB/caching/db.go:3446`).
- If close races with in-flight prep tasks, some tasks can remain unread in `l.vlogPrepCh` and never publish to `results`, leaving the caller blocked.
- Suggested fix: make workers drain queued prep tasks before honoring close, or make collector close-aware with a bounded cancellation path that cannot block indefinitely.

2) Autotune encode-cost under-accounting for worker-prepared frames
- Worker prep path hard-disables encode timing (`measureEncode := false`) (`TreeDB/caching/db.go:4429`).
- Later, `valueLogAutotuneMetrics.observe(...)` sees `encodeNsTotal` near zero in prepared-frame-heavy paths (`TreeDB/caching/db.go:4949`), biasing keep-policy to think encode is cheaper than it is.
- Suggested fix: account wall-time prep cost once per batch (not per frame) and feed a bounded estimate into autotune metrics.

3) K-boost unit tests don’t represent default runtime config
- Runtime default sets `valueLogDictMaxK` to 32 (`TreeDB/caching/db.go:2425`), but `chooseValueLogDictWriteK` boosts toward 64/96 (`TreeDB/caching/vlog_dict.go:555`) and tests use zero-value DB (`TreeDB/caching/vlog_dict_write_k_test.go:6`), so tests currently validate a path that default open won’t take.
- Suggested fix: add one integration-style test with normalized open defaults to verify intended behavior in production config.

PR #440, opened 2026-02-12
TreeDB random-read: append pointer Get path + snapshot reader reuse

snissn:
## Summary
- Route safe point reads (`Tree.Get`) for pointer-backed values through `ReadUnsafeAppend` when available, instead of `ReadUnsafe`.
- Cache the append-capable reader in `tree.Tree` during `New`/`Reset` to avoid repeated assertions.
- Reuse a snapshot-owned `valueReader` in `db.Snapshot` so `AcquireSnapshot` no longer boxes a fresh `valueReader{...}` per get.
- Clear the snapshot-owned reader on pool return.
- Add tests that lock in behavior:
  - `Tree.Get` uses append reader for pointer values.
  - `Tree.GetUnsafe` keeps using unsafe view reads.

## Why
Main-branch `unified-bench` random-read profiling (500k keys, fast profile) showed:
- `alloc_objects/random_read`: `DB.AcquireSnapshot` was the top allocator by count.
- Point reads were on the `ReadUnsafe -> readViaMmapView` path, not append-based pointer materialization.

These changes implement the planned step-2/step-3 improvements for the random-read path.

### Tests
- `go test ./TreeDB/tree`
- `go test ./TreeDB/db`

### Bench/profiles (same command as issue context)
Command:
`/usr/bin/time -p ./bin/unified-bench -dbs treedb -profile fast -keys 500000 -format markdown -checkpoint-between-tests -valsize 256 -dataset-val-pattern repeat_tail64 -val-pattern repeat_tail64 -profile-dir <OUT>`

Main baseline profile dir:
- `/tmp/gomap_main_profiles_hbK7wN`

This branch profile dir:
- `/tmp/gomap_branch_profiles_C4yLdL`

Observed random-read profile deltas:
- `alloc_objects` total: `1,003,277` -> `360,352` (about `-64%`)
- Top allocation path switched from snapshot setup to append read path:
  - before: `db.(*DB).AcquireSnapshot` dominated
  - after: `valuelog.(*File).readViaMmapAppend` + `valuelog.grow`
- CPU call path now shows append reads:
  - `db.valueReader.ReadUnsafeAppend`
  - `valuelog.(*File).readViaMmapAppend`
  - and no `readViaMmapView` in the hot top path for random-read.

Random-read ops/s remained in the same noisy band run-to-run for this benchmark shape.

### DIFF
--- a/TreeDB/tree/iterator.go
+++ b/TreeDB/tree/iterator.go
@@ -337,14 +337,6 @@ type Iterator struct {
 	verifyAlways   bool
 }
 
-type slabUnsafeAppender interface {
-	ReadUnsafeAppend(ptr page.ValuePtr, dst []byte) ([]byte, error)
-}
-
-type slabUnsafeBatchAppender interface {
-	ReadUnsafeAppendBatch(ptrs []page.ValuePtr, dst [][]byte) ([][]byte, error)
-}
-
 const iteratorPointerBatchMax = 2
 
 func (t *Tree) Iterator(start, end []byte) iterator.UnsafeIterator {
--- a/TreeDB/tree/tree.go
+++ b/TreeDB/tree/tree.go
@@ -33,6 +33,17 @@ type SlabReader interface {
 	ReadUnsafe(ptr page.ValuePtr) ([]byte, error)
 }
 
+// Optional fast path for append-style pointer reads that can reuse caller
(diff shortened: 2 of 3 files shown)


snissn:
Superseded by #444 (consolidated integration with follow-up Copilot remediations).

PR #482, opened 2026-02-14
valuelog: avoid allocations in mmap read path for destination-less appends

snissn:
## Summary
- Optimize `TreeDB/internal/valuelog/reader_mmap.go` in `readViaMmapAppend` for destination-less reads (`dst == nil`).
- Return zero-copy views from the mmap path for non-grouped and grouped-frame records when no append destination is supplied.
- Preserve the existing copy/append behavior when the caller provides `dst`.
- Avoid cache bookkeeping in non-append (`dst == nil`) paths to reduce lock overhead and copy-only-only work.

## Why
- Pointer-mode fast benchmark reads were still dominated by value-log read-path overhead.
- After this change, `random_read` in true-pointer mode improved versus earlier runs in this environment and allocation hotspots in `allocs_random_read_treedb` no longer show `readViaMmapAppend`-driven growth behavior.

## Benchmarks
- True-pointers benchmark command:
  - `./bin/unified-bench -dbs treedb -profile fast -test -keys 500000 -format markdown -checkpoint-between-tests -treedb-force-value-pointers=true -valsize=1 -profile-dir <profile-dir> -progress=false`
- False-pointers comparison run was also executed with identical flags for apples-to-apples checks.

## Notes
- All pre-existing, unrelated working-tree changes were stashed out before this PR to keep this change focused.

### DIFF
--- a/TreeDB/internal/valuelog/reader_mmap.go
+++ b/TreeDB/internal/valuelog/reader_mmap.go
@@ -2,7 +2,9 @@ package valuelog
 
 import (
 	"encoding/binary"
+	"reflect"
 	"runtime"
+	"unsafe"
 
 	"github.com/snissn/gomap/TreeDB/internal/crc"
 	"github.com/snissn/gomap/TreeDB/internal/limits"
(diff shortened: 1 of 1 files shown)


snissn:
## Issue #510 perf follow-up (KEYS=500000)

I re-ran the outerleaf V2 fast-mode perf check on this low-hanging read-path fix with deterministic input and focused tests.

### Before vs After (RUNS=1, WARMUP_RUNS=0, tests: `batch_write,batch_write_steady,prefix_scan,random_read,random_read_parallel,random_read_parallel_acquire_snapshot`)

| test | before (ops/s) | after (ops/s) | Δ vs before | v2_fenceptr / v1_forceptr ratio before | after |
| batch_write | 19,086,033 | 17,494,675 | -8.34% | 1.758x | 1.711x |
| batch_write_steady | 2,794,919 | 2,879,001 | +3.01% | 2.091x | 2.018x |
| prefix_scan | 6,501,951 | 6,610,499 | +1.67% | 0.407x | 0.425x |
| random_read | 6,188,396 | 6,144,346 | -0.71% | 1.163x | 1.158x |
| random_read_parallel | 42,299,824 | 44,579,835 | +5.39% | 1.202x | 1.317x |
| random_read_parallel_acquire_snapshot | 9,030,793 | 9,405,777 | +4.15% | 0.994x | 1.080x |

### Gate outcome
- RUNS=1 (strict gate) => PASS (`overall gate: PASS`).
- RUNS=3, WARMUP=1 => PASS except strict-noise fail on `prefix_scan` (`v1_inline cv 12.10%` vs `max 12.00%` threshold).

Overall this iteration did not introduce a major regression envelope; read-heavy modes stayed within expected range and random read regressions did not reappear.

PR #617, opened 2026-02-22
caching: fix pathological v2 fence promotion slowdown in batch_write_steady

snissn:
## Summary
This fixes a severe `Batch Write (Steady)` throughput collapse in mixed `unified-bench` runs under `index_outer_leaf_mode=v2_fenceptr`.

Root cause was repeated pending-mutation lookups over the same decoded outer-leaf sibling keys during fence anchor promotion/rematerialization.
In the hot path this amplified calls into memtable `GetEntry` (`sort.Search`/`bytes.Compare` heavy), making the benchmark appear hung.

## What changed
- Added per-decoded-outer-leaf pending lookup cache in `fenceAnchorPromoter`:
  - `keysByPtr` now stores `decodedFenceKeys` metadata (keys + pending lookup cache).
  - `lookupPendingMutation` memoizes lookup result per key index for each decoded outer leaf.
- Updated both promotion paths to use cached pending lookups:
  - `findPromotion`
  - `maybeRematerializeDeleteFallback`
- Replaced linear anchor scan with binary search in `findPromotion`.

## Correctness
Behavior is unchanged semantically: this is a caching/indexing optimization for already existing lookup logic.
Promotion/rematerialization decisions still rely on the same pending/snapshot conditions.

## Benchmark reproduction and results
Command (pathological mixed sequence):
```bash
./bin/unified-bench -dbs treedb -profile fast -keys 500000 -progress=false -format markdown -checkpoint-between-tests -valsize 1 -test sequential_write,random_write,dataset_write_random,dataset_write_sorted,batch_write,batch_write_steady
```

Before (base `0a6a66c4`):
- `Batch Write (Steady) / TreeDB = 20,186`

After (this PR):
- `Batch Write (Steady) / TreeDB = 351,416`

Improvement on this pathological sequence: ~17.4x.

Additional sanity:
- Full progression now moves past steady into subsequent tests (e.g. `Batch Random`) without apparent hang.
- `batch_write_steady` in isolation remains fast (`2,843,927` in local run).

## Validation
- `GOWORK=off go test ./TreeDB/caching -count=1`
- `GOWORK=off go test ./TreeDB/tree -count=1`
- `GOWORK=off go test ./TreeDB/db -count=1`

### DIFF
--- a/TreeDB/caching/db.go
+++ b/TreeDB/caching/db.go
@@ -2194,6 +2194,47 @@ func (p *fenceAnchorPromoter) maybePromoteAnchor(key []byte, pending fencePendin
 	return p.emitUniquePointer(promoteKey, entry.ValuePtr, emitPointer)
 }
 
+func (p *fenceAnchorPromoter) groupNeedsFullEmission(keys [][]byte, start, end int) (bool, error) {
+	if p == nil || start >= end || end > len(keys) {
+		return false, nil
+	}
+	p.loadBackendRange()
+	if p.rangeKnown {
+		if !p.backendRange.valid {
+			return false, nil
+		}
+		first := keys[start]
+		last := keys[end-1]
+		if len(first) == 0 || len(last) == 0 {
+			return true, nil
+		}
+		if bytes.Compare(last, p.backendRange.min) < 0 || bytes.Compare(first, p.backendRange.max) > 0 {
+			return false, nil
+		}
+	}
+
+	snap := p.snapshotView()
+	if snap == nil {
+		// If we cannot inspect exact entries, stay conservative.
+		return true, nil
+	}
+	for i := start; i < end; i++ {
+		key := keys[i]
+		if len(key) == 0 {
+			continue
+		}
+		_, err := snap.GetEntryExact(key)
+		if err == nil {
+			return true, nil
+		}
+		if errors.Is(err, tree.ErrKeyNotFound) {
+			continue
+		}
+		return false, err
+	}
+	return false, nil
+}
+
 func (p *fenceAnchorPromoter) loadBackendRange() {
 	if p == nil || p.rangeLoaded || p.db == nil {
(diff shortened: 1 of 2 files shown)


snissn:
Addressed the `dataset_write_sorted` fence-density regression in `f4c1516c`.

### Root cause
`emitWholeGroup` used a coarse range-overlap heuristic (`keyCouldExistInBackend`) and promoted full exact sibling emission for many disjoint sorted groups. That made `dataset_write_sorted` inflate index entries even when no exact-key overlap existed.

### Fix
- Added `groupNeedsFullEmission(...)` to `fenceAnchorPromoter`:
  - fast range prune (no backend overlap => no full emission)
  - exact overlap check via `Snapshot.GetEntryExact` across the source group
  - full emission only when exact entries actually exist (or snapshot unavailable -> conservative)
- Wired this into both grouped deferred pointer flush paths.
- Added regression test:
  - `TestCachingDB_FlushFenceModeCollapsedGroupSkipsExactEmissionWithoutExactOverlap`
  - ensures range-overlap-only does not emit dense exact siblings.

### Validation (same command shape)
`./bin/unified-bench -dbs treedb -profile fast -keys 500000 -progress=false -format markdown -checkpoint-between-tests -profile-dir=/tmp/fenceptr_dsfix_postfmt -vacuum-between-tests`

Before (reported):
- `Dataset Write (Sorted)`: `Before Test 8 MiB -> 6 MiB`
- `Before Batch Write`: `38 MiB -> 30 MiB`
- End `index.db`: `50 MiB`

After (`f4c1516c`):
- `Dataset Write (Sorted)`: `Before Test 4 MiB -> 2 MiB`
- `Before Batch Write`: `5 MiB -> 3 MiB`
- End `index.db`: `17 MiB`

Throughput remained healthy; batch/front-end write rates are comparable or better on this run shape.

PR #983, opened 2026-04-17
[codex] Use outer-leaf dicts on live leaf_vlog writes

snissn:
## What changed

This makes live split `leaf_vlog` writes eligible to use the dedicated `outer_leaf` dictionary path instead of always forcing block compression on the foreground write path.

The branch does three things:
- wires live `AppendLeafPage` writes to look up and pass the current `outer_leaf` dict ID when split outer-leaf dict mode is enabled
- teaches live leaf-lane auto/dict selection to prefer the dedicated outer-leaf dict once one exists
- exposes the reserved leaf-log lane in the cache-layer stats so live dict usage is actually visible in runtime metrics

It also updates the `fast` and `wal_on_fast` profiles to use split outer-leaf dict mode and adds end-to-end tests around live outer-leaf dict publication and selection.

## Why it changed

Recent Celestia control runs on `main` showed a large post-dwell to post-offline-rewrite size gap:
- `fast`: `2536171948 -> 2096276342` bytes (`-439895606`)
- `wal_on_fast`: `2585213462 -> 2117928675` bytes (`-467284787`)

The important finding was that this gap was not mainly a leaf-pack reclaim problem.

On both profiles, the largest pre-rewrite `leaf_vlog` files were:
- `grouped_k1`
- `dict_frames=0`
- `block_frames=all`
- `block_codec=snappy`

After offline rewrite on the same DBs, the largest rewritten `leaf_vlog` files became:
- `grouped_k1`
- `dict_frames=all`
- `block_frames=0`

So rewrite is already using dict compression for outer-leaf payloads, while the live dwell path was not. The root cause in the code was straightforward: live `AppendLeafPage` writes were always calling into the value-log writer with `dictID=0`, and the live selector still preferred block mode on the reserved leaf lane even after a dedicated outer-leaf dict existed.

## Impact

This is aimed at closing the dwell-to-rewrite size gap by making the live path use the same dedicated outer-leaf dict machinery that rewrite already benefits from.

It does not change the generic index outer-log format. It only changes the split `leaf_vlog` path for outer-leaf payloads.

## Validation

Local validation completed:
- `GOWORK=off go test ./TreeDB/caching -run 'Test(CachingLeafPageLog_AppendLeafPageUsesOuterLeafDictFramesInAutoBalancedSplitMode|ResolveVlogWriteMode_OuterLeafDictUsesDictOnLeafLaneInBalancedAuto|CachingLeafPageLog_AppendLeafPageCompactsSparseLeafPayload|ValueLogDictClassifierBypass_AllowsOuterLeafPagesForAutoBalancedSplitMode|ShouldBypassValueLogDictForRecords_OuterLeafPagesAllowedForAutoBalancedSplitMode)$' -count=1`
- `GOWORK=off go test ./TreeDB/caching -count=1`
- `GOWORK=off go test ./TreeDB -run 'Test(ApplyProfile_(FastSetsPolicyBools|WALOnFastKeepsWALOn)|ValueLogCompressionAuto_DefaultsToDictTrainingAndProbesDict|DictDB_DoesNotInheritOuterLeavesInValueLog)$' -count=1`

Control analysis on recent `main` runs is complete and is the reason for this PR.

A full live `run_celestia + 900s dwell` validation from this branch is currently running and will be attached to this PR once it finishes.

### DIFF
--- a/TreeDB/caching/db.go
+++ b/TreeDB/caching/db.go
@@ -24784,6 +24784,13 @@ func (db *DB) Stats() map[string]string {
 	stats["treedb.cache.vlog_dict.last_publish_unix_nano"] = fmt.Sprintf("%d", db.valueLogDictLastPublishUnixNano.Load())
 	stats["treedb.cache.vlog_dict.last_k_update_unix_nano"] = fmt.Sprintf("%d", db.valueLogDictLastKUpdateUnixNano.Load())
 	stats["treedb.cache.vlog_dict.current_k"] = fmt.Sprintf("%d", db.valueLogDictCurrentK.Load())
+	for class := 0; class < vlogDictClassCount; class++ {
+		suffix := vlogDictClassSuffix(vlogDictClass(class))
+		stats["treedb.cache.vlog_dict.last_applied_dict_id."+suffix] = fmt.Sprintf("%d", db.valueLogDictLastAppliedDictIDByClass[class].Load())
+		stats["treedb.cache.vlog_dict.last_applied_dict_hash."+suffix] = fmt.Sprintf("%x", db.valueLogDictLastAppliedDictHashByClass[class].Load())
+		stats["treedb.cache.vlog_dict.current_k."+suffix] = fmt.Sprintf("%d", db.valueLogDictCurrentKByClass[class].Load())
+		stats["treedb.cache.vlog_dict.cached_current_id."+suffix] = fmt.Sprintf("%d", db.dictCurrentCachedByClass[class].Load())
+	}
 	db.valueLogDictBytesMu.Lock()
 	stats["treedb.cache.vlog_dict.cached_dict_id"] = fmt.Sprintf("%d", db.valueLogDictBytesID)
 	stats["treedb.cache.vlog_dict.cached_dict_bytes"] = fmt.Sprintf("%d", len(db.valueLogDictBytes))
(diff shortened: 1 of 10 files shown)


snissn:
Live `run_celestia + 900s dwell` validation from this branch against `wal_on_fast` is complete.

Result: the branch does fix the intended runtime behavior, but it is not merge-ready as-is.

What worked:
- live outer-leaf dict frames are now active during the real run
- the run entered dwell cleanly and completed successfully
- final dwell-end size is slightly better than the recent `main wal_on_fast` control

Measured candidate result:
- sync: `448s`
- total: `1348s`
- max RSS: `13,561,808 kB`
- final `application.db`: `2,576,203,003 B`
- final `leaf_vlog`: `2,344,984,993 B`
- final `value_vlog`: `169,695,667 B`

Recent `main wal_on_fast` control for comparison:
- sync: `326s`
- total: `1226s`
- max RSS: `10,886,772 kB`
- final `application.db`: `2,585,256,162 B`
- final `leaf_vlog`: `2,351,823,540 B`
- final `value_vlog`: `171,297,961 B`

Delta vs control:
- `application.db`: better by only `9,053,159 B`
- `leaf_vlog`: better by `6,838,547 B`
- `value_vlog`: better by `1,602,294 B`
- but sync regressed by `122s`
- and max RSS regressed by about `2.67 GiB`

Control-side rewrite floor on the same profile is still much lower:
- post-dwell `main wal_on_fast`: `2,585,213,462 B`
- post-offline-rewrite: `2,117,928,675 B`
- remaining gap on `main`: about `467 MB`

So this branch proves the root cause: live dwell was missing outer-leaf dict compression that rewrite already uses. But the current implementation is too expensive in production terms relative to the size win it produces.

I am leaving this PR in draft. The next step should be to reduce the cost of live outer-leaf dict use, likely by making it much more selective or by shifting more of this benefit into maintenance/rewrite instead of the hot restore path.

snissn:
Closing this experiment.

Reason:
- recent measurement shows this line of outer-leaf optimization is not a production win relative to `main`
- the useful conclusions and benchmark notes are preserved in the branch/PR history
- current `main` plus offline `vlog-rewrite` remains the better operational path for now

Status at close:
- this PR stack produced useful research, but not a mergeable maintenance strategy
- we are pausing this optimization line rather than carrying a stale open PR chain

snissn:
Closing experimental outer-leaf optimization PR; measured result is not better than main.

PR #1011, opened 2026-04-22
[codex] valuelog: reuse full-frame decode for grouped reads

snissn:
## What changed
- reuse a direct owned full-frame decode for grouped compressed reads when the requested subrecord spans the entire decoded frame
- add a correctness regression for `ReadAtWithDictTo(..., nil)` on the whole-frame grouped-compressed path
- add a focused benchmark for the same whole-frame fallback read path

## Why
The grouped-compressed fallback in `decodeRecordTo` was still doing:
1. decode the full frame into scratch
2. allocate a second exact-sized slice
3. copy the full decoded payload into that second slice

That extra allocation and copy were unnecessary when the requested subrecord already covered the entire decoded frame. In that case we can return an owned full-frame decode directly.

## Impact
This is a narrow lower-level value-log reader optimization. It does not change on-disk format or read semantics.

Observed local benchmark signal on top of the current draft perf stack:
- `BenchmarkSnapshotValueLogGet/Get`: `6012 ns/op` -> `4933 ns/op`
- `BenchmarkSnapshotValueLogGet/GetAppend`: `5099 ns/op` -> `4712 ns/op`
- `BenchmarkDBValueLogGet/Get`: `8556 ns/op, 369 B/op` -> `5435 ns/op, 349 B/op`
- `BenchmarkDBValueLogGet/GetAppend`: `8723 ns/op` -> `6763 ns/op`
- `BenchmarkFileReadAppendCompressedWholeFrame/nil_dst`: `3221 ns/op, 4105 B/op, 1 alloc/op` on the new focused path bench

## Validation
- `GOWORK=off go test ./TreeDB/internal/valuelog -count=1`
- `GOWORK=off go test ./TreeDB/internal/valuelog -run 'TestReadAtWithDictTo_(UsesDstForCompressedGroupedSubrecord|CompressedWholeFrameNilDst)$' -count=1`
- `GOWORK=off go test ./TreeDB/internal/valuelog -run '^$' -bench 'BenchmarkFileReadAppendCompressedWholeFrame/nil_dst$' -benchmem -benchtime=2s -count=1`
- `GOWORK=off go test ./TreeDB -run '^$' -bench 'Benchmark(DB|Snapshot)ValueLogGet/(Get|GetAppend)$' -benchmem -benchtime=2s -count=1`

### DIFF
--- a/TreeDB/README.md
+++ b/TreeDB/README.md
@@ -112,6 +112,10 @@ through `wal/` even when WAL is off.
 
 Details: `docs/TREEDB_WRITE_PATHS.md`.
 
+If you are integrating TreeDB behind a Cosmos/Comet-style wrapper, use
+`TreeDB/integration/kvstoreadapter` to standardize profile/env/default handling
+instead of re-copying the open-path glue in each downstream repo.
+
 ## Leaf Pages in the Value Log
 
 `Options.IndexOuterLeavesInValueLog` stores the B+Tree **leaf pages** (the 4096B
--- a/TreeDB/README.md
+++ b/TreeDB/README.md
@@ -102,8 +102,8 @@ db, err := treedb.Open(opts)
 Profiles are intended to make intent explicit:
 
 - `ProfileDurable`: safest defaults (recommended).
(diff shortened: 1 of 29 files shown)


snissn:
Closing this for now. Rebuilt clean on top of #1009 and reran the focused DB/Snapshot benches; the result was mixed, not a clear standalone perf win. Keeping the local worktree for follow-up if we want to revisit this path later.

PR #1195, opened 2026-05-02
Copy unchanged leaf-log internal refs in zipper

snissn:
## Summary

Part of #1159's deeper zipper / leaf-log root-apply cost track.

This PR adds a narrow fast path for rebuilding internal pages whose children are value-log-backed outer leaves:

- adds `node.Builder.AddInternalLeafLogChildFromNode` to copy an already encoded leaf-log child entry from a source internal node
- uses that path in zipper internal merges when a child has no ops and therefore its child ref/key are unchanged
- preserves existing zipper internal child-ref metrics when the copy path is used
- adds node tests and a focused microbenchmark for the copied-entry primitive

This does not reduce root applies, leaf-log reads, or leaf-log pages written. It only makes one part of unavoidable internal child construction cheaper when unchanged leaf-log child entries are path-copied during root apply.

## Validation

```bash
go test ./TreeDB/node ./TreeDB/zipper -count=1

go test ./TreeDB/db -count=1

go test ./TreeDB/node -run '^$' -bench '^BenchmarkBuilderAddInternalLeafLogChildRef$' -benchmem -count=3

MONGO_GATEWAY_PROFILE_BENCH_UPDATE_DOCUMENTS=100000 \
go test ./cmd/mongo_gateway_bench \
  -run '^$' \
  -bench '^BenchmarkDirectCollectionConcurrentUpdateBSONIndexes2$' \
  -benchtime=100000x \
  -benchmem \
  -count=2

git diff --check
```

## Focused primitive benchmark

```text
BenchmarkBuilderAddInternalLeafLogChildRef/decode_reencode-8  1771 ns/op  4928 B/op  3 allocs/op
BenchmarkBuilderAddInternalLeafLogChildRef/decode_reencode-8  1743 ns/op  4928 B/op  3 allocs/op
BenchmarkBuilderAddInternalLeafLogChildRef/decode_reencode-8  1760 ns/op  4928 B/op  3 allocs/op
BenchmarkBuilderAddInternalLeafLogChildRef/copy_encoded-8     1201 ns/op  4928 B/op  3 allocs/op
BenchmarkBuilderAddInternalLeafLogChildRef/copy_encoded-8     1200 ns/op  4928 B/op  3 allocs/op
BenchmarkBuilderAddInternalLeafLogChildRef/copy_encoded-8     1200 ns/op  4928 B/op  3 allocs/op
```

Interpretation: the primitive is roughly 30% faster for rebuilding a 96-entry internal node with unchanged leaf-log child refs. Allocation count is unchanged because the benchmark allocates a page buffer per rebuild in both variants.

## Macro smoke

The direct BSON two-index update macro row was noisy/neutral rather than a clear end-to-end win:

```text
100000  5023 ns/op  199098 docs/sec  1124 B/op  2 allocs/op
100000  4895 ns/op  204293 docs/sec  1132 B/op  2 allocs/op
```

Interpretation: this should be reviewed as a low-risk CPU cleanup for one zipper substep, not as a standalone throughput breakthrough. The larger #1159 wins still need to come from reducing leaf-log pages read/written/rebuilt, current-read cost, and/or root-apply frequency.

### DIFF
--- a/TreeDB/node/builder.go
+++ b/TreeDB/node/builder.go
@@ -887,6 +887,50 @@ func (b *Builder) addInternalLeafLogChild(key []byte, ref page.LogRecordRef) err
 	return nil
 }
 
+// AddInternalLeafLogChildFromNode appends an already-encoded leaf-log child
+// entry from src. Zipper path-copy rebuilds use this when an internal child is
+// unchanged, avoiding LogRecordRef decode/re-encode work while preserving the
+// encoded on-page representation exactly.
+func (b *Builder) AddInternalLeafLogChildFromNode(src *Node, index uint16) error {
+	if b.pType != page.PageTypeInternal || src == nil || !src.internalLeafLogRefs() {
+		return ErrInvalidType
+	}
+	if b.internalBaseDelta {
+		return ErrInvalidType
+	}
+	if b.count > 0 && !b.internalLeafLogRefs {
+		return ErrInvalidType
+	}
+
+	offset, err := src.getOffset(index)
+	if err != nil {
+		return err
+	}
+	ptr := int(offset)
+	if ptr+internalLeafLogRefHeaderSize > len(src.data) {
+		return ErrCorruptedNode
+	}
+	keyLen := int(binary.LittleEndian.Uint16(src.data[ptr : ptr+2]))
+	entrySize := internalLeafLogRefHeaderSize + keyLen
+	if ptr+entrySize > len(src.data) {
+		return ErrCorruptedNode
+	}
+	required := entrySize + DirectoryEntrySize
+	if b.heapStart-b.dirEnd < required {
+		return ErrNodeFull
+	}
+
+	entryStart := b.heapStart - entrySize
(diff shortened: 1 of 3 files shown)


snissn:
CI fix pushed in 75f549cb23: the benchmark now uses the Go 1.23-compatible `for n := 0; n < b.N; n++` loop instead of `b.Loop()`.

Local validation after the fix:

```text
go test ./TreeDB/node ./TreeDB/zipper -count=1
# ok github.com/snissn/gomap/TreeDB/node
# ok github.com/snissn/gomap/TreeDB/zipper

go test ./TreeDB/node -run ^\$ -bench ^BenchmarkBuilderAddInternalLeafLogChildRef\$ -benchmem -count=1
BenchmarkBuilderAddInternalLeafLogChildRef/decode_reencode-8  718358  1663 ns/op  4096 B/op  1 allocs/op
BenchmarkBuilderAddInternalLeafLogChildRef/copy_encoded-8    1000000 1134 ns/op  4096 B/op  1 allocs/op
```

PR #1225, opened 2026-05-03
collections: add opt-in indexed flush overlay roots

snissn:
## Summary
- add opt-in `BufferedIndexedOverlayRoots` for indexed collection memtable flushes
- publish flushed indexed root runs as durable `collections/root-overlay/<root>` descriptors instead of applying them into base roots
- allow indexed buffered inserts/updates to read through existing overlays and keep non-overlay writes guarded
- make insert duplicate/unique preflight overlay-aware
- coalesce the normal single-overlay case by applying each new overlay delta against the previous overlay root, so durable overlay reads do not grow an unbounded root stack
- add benchmark env support via `MONGO_GATEWAY_PROFILE_BENCH_BUFFERED_INDEXED_OVERLAY_ROOTS=true`

This is intentionally architecture-only and opt-in. It is not a default speed win yet; it proves the committed overlay-root write-back shape and exposes the next architectural bottleneck clearly. The latest benchmark shows that coalescing fixes the worst read fanout from the naive overlay path, but root apply work still re-enters the publish path when we coalesce by applying against the prior overlay root.

## Validation
- `GOWORK=off go test ./TreeDB/collections -run 'TestCollectionIndexedOverlayRootFlushSupportsReadsUpdatesAndUniqueChecks|TestCollectionCatalogOverlayRootsAreVisibleAfterReopen' -count 1`
- `GOWORK=off go test ./TreeDB/collections -count 1`
- `GOWORK=off go test ./TreeDB/db -count 1`
- `GOWORK=off go test ./cmd/mongo_gateway_bench -count 1`
- `GOWORK=off go test ./cmd/mongo_gateway_bench -run 'TestProfileBenchBufferedIndexedAsyncFlushEnv' -count 1`
- `GOWORK=off go test ./TreeDB/collections -run 'TestCollectionIndexedOverlayRootFlushSupportsReadsUpdatesAndUniqueChecks' -count 1`
- `git diff --check`

A broad `GOWORK=off go test ./TreeDB/... -count 1` had one transient unrelated failure in `TreeDB/caching`: `TestVlogGenerationChunkHarness_StagesAndExecutesRealChunkDebt`, output `remaining chunk live bytes=12420 want < initial 12420 (chunks before=3 after=3)`. Rerunning that exact test and then `GOWORK=off go test ./TreeDB/caching -count 1` both passed.

## Benchmark notes
Baseline/default path, overlay roots disabled:

```bash
[...]

PR #1226, opened 2026-05-03
collections: compact overlay roots into base roots

snissn:
## Summary
- add `Collection.CompactRootOverlays(ctx)` as the explicit maintenance boundary for durable collection overlay roots
- fold each overlay root stack into a rebuilt base collection root and clear the corresponding `collections/root-overlay/<root>` descriptor in the same backend commit
- preserve visible primary/secondary/index-state semantics by compacting through the catalog-root merged iterator, so overlay tombstones still hide base entries
- refresh collection and write-domain catalog caches after compaction
- add `MONGO_GATEWAY_PROFILE_BENCH_COMPACT_OVERLAY_ROOTS_AFTER_FLUSH=true` so the focused mongo gateway update benchmark can compact overlays after preload/warmup flushes before the timed phase

This is the stacked architectural follow-up to #1225. #1225 proves durable overlay roots as a committed write-back shape; this PR adds the first explicit mechanism to return those overlays to a simple one-root read shape. It is still an explicit maintenance primitive, not the final background scheduler.

## Validation
- `GOWORK=off go test ./TreeDB/collections -run 'TestCollectionCompactRootOverlaysFoldsIntoBaseRoots|TestCollectionIndexedOverlayRootFlushSupportsReadsUpdatesAndUniqueChecks|TestCollectionIndexedOverlayRootValidationRejectsStaleOverlayDescriptors' -count 1`
- `GOWORK=off go test ./TreeDB/collections -count 1`
- `GOWORK=off go test ./TreeDB/db -count 1`
- `GOWORK=off go test ./cmd/mongo_gateway_bench -run 'TestProfileBenchBufferedIndexedAsyncFlushEnv' -count 1`
- `GOWORK=off go test ./cmd/mongo_gateway_bench -count 1`
- `git diff --check`

## Focused benchmark
Command:

```bash
MONGO_GATEWAY_PROFILE_BENCH_UPDATE_DOCUMENTS=1000000 \
MONGO_GATEWAY_PROFILE_BENCH_BUFFERED_INDEXED_OVERLAY_ROOTS=true \
MONGO_GATEWAY_PROFILE_BENCH_COMPACT_OVERLAY_ROOTS_AFTER_FLUSH=true \
GOWORK=off go test ./cmd/mongo_gateway_bench \
  -run '^$' \
  -bench '^BenchmarkDirectCollectionConcurrentUpdateBSONIndexes3CityUpdate$' \
  -benchtime=1000000x \
  -benchmem \
  -count=1
```

Result: `11575 ns/op`, `86393 docs/sec`, `13953 B/op`, `10 allocs/op`; `compact_overlay_roots_after_flush=1`, `preload_overlay_compact_roots=4`, `warmup_overlay_compact_roots=2`, `update_current_read_ns/doc=5673`, `publish_delta_group_root_apply_ns/doc=3696`.

Interpretation: this is a correctness/maintenance primitive, not the throughput win by itself. It confirms that cadence alone does not resolve the hot-path architecture: even after compaction before the timed phase, coalesced overlay flushes still pay root apply, and current reads still remain too expensive. The next high-value architectural work should avoid per-flush root apply and/or avoid overlay miss probes, likely through background/prebuilt root apply, overlay read manifests/filters, or a committed overlay read index.

## Architectural note
This intentionally does not claim to solve the hot publish path by itself. It gives the overlay-root design a correctness-tested collapse operation that future PRs can schedule asynchronously or replace with prebuilt/off-lock root application. The expected end state remains: writes publish durable flush units cheaply, compaction folds overlays before read fanout grows, and eventually root construction moves out of the writer critical path.

### DIFF
--- a/TreeDB/collections/api.go
+++ b/TreeDB/collections/api.go
@@ -538,6 +538,11 @@ type CollectionOptions struct {
 	// Flush, FlushAll, and backend Close still drain pending indexed writes before
 	// returning, but auto-flush thresholds no longer imply immediate durability.
 	BufferedIndexedAsyncFlush bool `json:"buffered_indexed_async_flush,omitempty"`
+	// BufferedIndexedOverlayRoots publishes indexed memtable flush units as
+	// durable overlay roots instead of immediately applying them into base roots.
+	// Maintenance can later compact overlay roots into base roots; reads merge
+	// overlay roots over base roots while they are pending.
+	BufferedIndexedOverlayRoots bool `json:"buffered_indexed_overlay_roots,omitempty"`
 	// BufferedIndexedAsyncFlushMaxQueuedUnits bounds immutable indexed flush
 	// units queued for the background publisher. Zero uses the native default
 	// when async flush is enabled on an indexed schema.
@@ -611,6 +616,7 @@ type indexedFlushPublishWork struct {
 	flushUnit      indexedFlushUnit
 	rootNames      []string
 	rootBaseIDs    map[string]uint64
+	rootOverlays   map[string][]uint64
 	docCount       int
 	byteCount      int64
 	rootRunCount   int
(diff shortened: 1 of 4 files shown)


snissn:
@codex review

snissn:
Correction to the previous note: addressed the Copilot review in 8b60b43fb2. Overlay compaction still rebuilds touched collection roots, but descriptor updates now use `PublishOrderedRootDeltaGroupWithSystemDeltaBuilder` plus `buildSystemDeltaIterator`, so the system-root update is a small touched-key delta rather than a full system-root materialization.

PR #1234, opened 2026-05-03
db: parallelize eligible mixed root applies

snissn:
## Summary

- allow ordered-root batch groups with mixed opt-in roots to parallelize the eligible roots when at least two active roots are marked `ParallelApply`
- preserve serialized behavior when fewer than two roots opt in
- add coverage for a three-root mixed group: two parallel-safe roots plus one serialized root, with correctness checks and root-apply counter assertions

## Why this is architectural

The previous batch-publish rule required every active root in a group to opt into parallel apply. That made one conservative root force the whole group through the serialized apply path. This PR keeps per-root opt-in semantics but removes the whole-group coupling, which is the publish fan-in shape needed for future collection root mixes.

## Validation

```bash
go test ./TreeDB/db -run 'TestPublishOrderedRootDeltaBatchGroupWithSystemDeltaBuilder_Optimistic(MixedOptInParallelizesEligibleRoots|MixedOptInStaysSerialized|ColdBuildsRootsInParallel|AppliesRootsInParallel|PreservesConcurrentUserRootWrite|RebasesSystemDeltaAfterConcurrentSystemCommit)' -count=1
go test ./TreeDB/db -count=1
go test ./cmd/mongo_gateway_bench -count=1
```

The new focused test verifies `parallel_groups=1`, `parallel_roots=2`, and `root_apply_calls=3` for a mixed three-root publish. The existing one-opt-in mixed test still verifies `parallel_groups=0`, so the minimum parallel-root threshold is unchanged.

Stacked on #1233. Refs #1159.

### DIFF
--- a/TreeDB/db/ordered_root_publish.go
+++ b/TreeDB/db/ordered_root_publish.go
@@ -1401,18 +1401,16 @@ func (db *DB) publishOrderedRootDeltaBatchGroupWithSystemDeltaBuilder(ordered []
 }
 
 func orderedRootDeltaBatchGroupParallelApplyEligible(ordered []OrderedRootDeltaBatchPublishInput) bool {
-	active := 0
 	parallelActive := 0
 	for idx := range ordered {
 		if ordered[idx].Delta == nil || ordered[idx].Delta.IsEmpty() {
 			continue
 		}
-		active++
 		if ordered[idx].ParallelApply {
 			parallelActive++
 		}
 	}
-	return active == parallelActive && parallelActive >= orderedRootDeltaBatchGroupParallelApplyMinRoots
+	return parallelActive >= orderedRootDeltaBatchGroupParallelApplyMinRoots
 }
 
 func (db *DB) applyOrderedRootDeltaBatchGroupRoots(idx *indexGen, ordered []OrderedRootDeltaBatchPublishInput, alloc zipper.PageAllocator, coldBuildAlloc bulk.Allocator) ([]orderedRootDeltaBatchGroupApplyResult, bool) {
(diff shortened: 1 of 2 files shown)


snissn:
Addressed Copilot review in b623cf8456. The mixed path now starts the opted-in parallel roots first, waits for them, and only then applies non-opted roots serially. That preserves the `ParallelApply` contract for conservative roots while avoiding the previous start-order coupling where a non-opted root completed before eligible roots could begin. Re-ran:

```bash
go test ./TreeDB/db -run 'TestPublishOrderedRootDeltaBatchGroupWithSystemDeltaBuilder_Optimistic(MixedOptInParallelizesEligibleRoots|MixedOptInStaysSerialized|ColdBuildsRootsInParallel|AppliesRootsInParallel)' -count=1
```

snissn:
Added a focused scheduler regression test in 78f455aed3 for the remaining Copilot concern. The test calls the grouped-root apply scheduler directly with two parallel cold roots and one serialized warm root, then uses separate allocator wrappers to fail if the serialized root allocator starts before both parallel roots have begun. This covers the start-order coupling without adding production hooks or hot-path instrumentation.\n\nRe-ran:\n\n```bash\ngo test ./TreeDB/db -run 'TestApplyOrderedRootDeltaBatchGroupRoots_MixedOptInStartsParallelBeforeSerial|TestPublishOrderedRootDeltaBatchGroupWithSystemDeltaBuilder_Optimistic(MixedOptInParallelizesEligibleRoots|MixedOptInStaysSerialized|ColdBuildsRootsInParallel|AppliesRootsInParallel)' -count=1\n```

PR #1246, opened 2026-05-04
collections: prove index minimization beyond mask width

snissn:
## Summary

Starts #1242 PR 2 as a behavior-preserving audit/proof slice for affected-index minimization:

- adds a 72-index `UpdateBatch` matrix for ordinals 63, 64, and 71;
- proves only the targeted secondary root changes when an indexed value changes at or beyond the `uint64` changed-mask boundary;
- verifies aggregate changed/unchanged index stats for the >64 fallback path;
- proves a unique index at ordinal 64 is skipped when unchanged, declines the buffered/no-secondary-unique path when changed, and still detects durable unique conflicts in normal `UpdateBatch`.

## Tests

```sh
```

### DIFF
--- a/TreeDB/collections/api_test.go
+++ b/TreeDB/collections/api_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"reflect"
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


snissn:

PR #1247, opened 2026-05-04
collections: prove direct buffered secondary skips

snissn:
## Summary

Continues #1242 PR 2 as another behavior-preserving audit/proof slice:

- adds direct-buffered BSON coverage for non-indexed updates where effective secondary values are unchanged;
- proves the BSON path stages only the primary root and does not stage unique/non-unique secondary root runs;
- adds direct-buffered template-v1 coverage where template shape changes but indexed values are unchanged;
- proves template-v1 stages primary/template roots separately while still skipping unchanged secondary roots.

## Tests

```sh
```

### DIFF
--- a/TreeDB/collections/update_minimization_direct_buffered_test.go
+++ b/TreeDB/collections/update_minimization_direct_buffered_test.go
@@ -1,0 +1,210 @@
+package collections
+
+import (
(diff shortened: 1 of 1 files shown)


snissn:

PR #1248, opened 2026-05-04
collections: prove direct BSON id mutation rejection

snissn:
## Summary

Adds another #1242 PR 2 audit/proof test for BSON replacement safety:

- covers the direct `Update` path, complementing existing `UpdateBatch` BSON `_id` mutation coverage;
- proves a BSON replacement that mutates `_id` is rejected;
- asserts the rejection does not advance commit sequence, does not change the primary root descriptor, and leaves the original document intact.

## Tests

```sh
```


snissn:

PR #1249, opened 2026-05-04
collections: pin no-index update publish baseline

snissn:
## Summary

Adds a #1242 PR 1a baseline test for source-confirmed no-index update behavior:

- proves no-index `Update` flushes/drains collection-local pending state and publishes synchronously;
- asserts the update advances commit sequence and changes the primary root descriptor immediately;
- asserts it does not create queued/publishing indexed flush units or indexed root runs;
- documents the current baseline so later no-index async write-back/coalescing can be reviewed as an explicit behavior change.

## Tests

```sh
```

### DIFF
--- a/TreeDB/collections/no_index_update_baseline_test.go
+++ b/TreeDB/collections/no_index_update_baseline_test.go
@@ -1,0 +1,100 @@
+package collections
+
+import (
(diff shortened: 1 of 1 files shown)


snissn:

PR #1251, opened 2026-05-04
collections: prove direct update index minimization

snissn:
## Summary

Adds another #1242 PR 2 audit/proof slice for direct update minimization parity:

- creates a 16-secondary-index collection on the direct update path;
- proves changing one indexed field changes only the primary root and the one affected secondary root;
- proves changing only a non-indexed field changes the primary root and skips every secondary root;
- verifies old/new secondary index lookup behavior for the changed target index.

## Tests

```sh
```


snissn:

PR #1252, opened 2026-05-04
collections: prove batch skips many unchanged indexes

snissn:
## Summary

Adds another #1242 PR 2 audit/proof test for `UpdateBatch` minimization:

- creates a 32-secondary-index collection on the non-buffered `UpdateBatch` path;
- replaces a document while changing only a non-indexed field;
- verifies changed/unchanged index stats report 0 changed and 32 unchanged;
- proves no secondary delete/set/key-byte work is emitted;
- proves every secondary root descriptor remains unchanged while the primary root changes.

## Tests

```sh
```

### DIFF
--- a/TreeDB/collections/update_batch_many_unchanged_test.go
+++ b/TreeDB/collections/update_batch_many_unchanged_test.go
@@ -1,0 +1,146 @@
+package collections
+
+import (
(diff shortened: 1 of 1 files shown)


snissn:

PR #1255, opened 2026-05-04
collections: prove BSON extraction minimization

snissn:
## Summary

Adds a focused #1242 PR 2 proof test for BSON extraction minimization:

- creates a BSON collection with a unique dotted-path `profile.email` index and a multikey `tags` array index;
- verifies a replacement that changes the BSON document but keeps effective dotted/array index values equal changes only the primary root;
- verifies changing only the dotted email root leaves the array secondary root unchanged;
- verifies changing only the multikey tag root leaves the dotted email secondary root unchanged;
- checks exact index lookup results throughout so the root-ID assertions are tied to visible secondary behavior.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionBSONUpdateSkipsUnchangedDottedAndArraySecondaryRoots' -count=1
```

### DIFF
--- a/TreeDB/collections/bson_extraction_minimization_test.go
+++ b/TreeDB/collections/bson_extraction_minimization_test.go
@@ -1,0 +1,171 @@
+package collections
+
+import (
+	"bytes"
+	"testing"
+
+	backenddb "github.com/snissn/gomap/TreeDB/db"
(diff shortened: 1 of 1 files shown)


snissn:

PR #1256, opened 2026-05-04
collections: prove BSON null index minimization

snissn:
## Summary

Adds another focused #1242 PR 2 BSON minimization proof:

- covers BSON missing and explicit `null` values for an indexed scalar field as the same empty effective value set;
- verifies missing -> null changes only the primary root and leaves secondary roots unchanged;
- verifies null -> indexed value changes only the target secondary root;
- verifies indexed value -> missing removes lookup visibility without changing unrelated secondary roots.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionBSONUpdateTreatsMissingAndNullIndexValuesAsUnchanged' -count=1
```

### DIFF
--- a/TreeDB/collections/bson_null_missing_minimization_test.go
+++ b/TreeDB/collections/bson_null_missing_minimization_test.go
@@ -1,0 +1,170 @@
+package collections
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


snissn:

PR #1257, opened 2026-05-04
collections: prove template-v1 root minimization

snissn:
## Summary

Adds a focused #1242 PR 2 template-v1 minimization proof:

- verifies a template-v1 replacement with the same template shape and unchanged indexed values changes only the primary root;
- verifies a replacement with a new template shape changes the template root while keeping secondary roots unchanged when indexed values are equal;
- verifies a city-only indexed update changes only the city secondary root and leaves the email secondary and template roots unchanged;
- checks exact index lookup visibility after each step.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionTemplateV1UpdateSeparatesTemplateAndSecondaryRootWork' -count=1
```

### DIFF
--- a/TreeDB/collections/template_v1_minimization_test.go
+++ b/TreeDB/collections/template_v1_minimization_test.go
@@ -1,0 +1,174 @@
+package collections
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


snissn:

PR #1263, opened 2026-05-04
collections: prove JSON null index minimization

snissn:
## Summary

Adds a focused #1242 PR 2 JSON/default null/missing minimization proof:

- verifies missing -> explicit `null` for an indexed scalar field changes only the primary root;
- verifies the persisted index-state root and absent secondary root stay unchanged for the missing/null empty effective value set;
- verifies null/missing -> indexed value updates the index-state and target secondary root without changing unrelated secondary roots;
- verifies indexed value -> missing removes lookup visibility and updates only the target index-state/secondary work.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionJSONUpdateTreatsMissingAndNullIndexValuesAsUnchanged' -count=1
```

### DIFF
--- a/TreeDB/collections/json_null_missing_minimization_test.go
+++ b/TreeDB/collections/json_null_missing_minimization_test.go
@@ -1,0 +1,170 @@
+package collections
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


snissn:

PR #1265, opened 2026-05-04
collections: prove BSON update batch null minimization

snissn:
## Summary

Adds a focused #1242 PR 2 BSON `UpdateBatch` null/missing minimization proof:

- verifies missing -> explicit BSON `null` for an indexed scalar field emits no secondary work and leaves the absent secondary root unchanged;
- asserts `LastUpdateStats` reports unchanged index work for the missing/null empty effective value set;
- verifies null/missing -> indexed value updates only the target secondary root and leaves unrelated secondary roots unchanged;
- checks exact secondary lookup visibility throughout.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionBSONUpdateBatchTreatsMissingAndNullIndexValuesAsUnchanged' -count=1
```

### DIFF
--- a/TreeDB/collections/bson_update_batch_null_minimization_test.go
+++ b/TreeDB/collections/bson_update_batch_null_minimization_test.go
@@ -1,0 +1,181 @@
+package collections
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


snissn:

PR #1267, opened 2026-05-04
collections: prove template-v1 update batch null minimization

snissn:
## Summary

Adds a focused #1242 PR 2 template-v1 `UpdateBatch` null/missing minimization proof:

- verifies missing -> explicit `null` for an indexed scalar field may update the template root, but emits no secondary work because the effective indexed value set remains empty;
- asserts `LastUpdateStats` reports unchanged index work for that missing/null update;
- verifies null/missing -> indexed value updates only the target secondary root while keeping the same template shape and unrelated secondary roots stable;
- checks exact secondary lookup visibility throughout.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionTemplateV1UpdateBatchTreatsMissingAndNullIndexValuesAsUnchanged' -count=1
```

### DIFF
--- a/TreeDB/collections/template_v1_update_batch_null_minimization_test.go
+++ b/TreeDB/collections/template_v1_update_batch_null_minimization_test.go
@@ -1,0 +1,186 @@
+package collections
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


snissn:

PR #1274, opened 2026-05-04
collections: populate direct update stats

snissn:
## Summary

Implements a small #1242 PR 2 observability tightening for the direct single-document `Update` path:

- populates `CollectionUpdateStats` for successful direct update outcomes, including matched/no-op/missing cases;
- records changed/unchanged index decisions, unique check/skip counts, secondary delete/set/key-byte work, secondary-run stats, and detailed per-index stats when enabled;
- keeps update semantics unchanged and reuses the existing stats shape instead of adding a broad new metrics subsystem.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionDirectUpdate.*Stats' -count=1
```

### DIFF
--- a/TreeDB/collections/api.go
+++ b/TreeDB/collections/api.go
@@ -359,9 +359,9 @@ type CollectionUpdateStats struct {
 	IndexValueUnchanged    int
 	UniqueIndexChecks      int
 	UniqueIndexCheckSkips  int
-	// IndexStats is populated only when detailed update-batch stats are enabled
-	// and the collection has at most len(IndexStats) index runtimes. The first
-	// IndexStatsCount entries are valid.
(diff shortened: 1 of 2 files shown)


snissn:
Follow-up pushed in f5c3fc53c2: detailed per-index update stats now populate up to the inline cap instead of disappearing for wider collections, and the helper is shared by direct Update and UpdateBatch planning. Added coverage for a direct update with more indexes than the inline stats cap.\n\nI did not move callback execution ahead of old-state extraction in this PR. That extraction order is pre-existing behavior on the direct path and changing it is closer to the #1242 2A current-state/materialization optimization bucket; this PR is scoped to observability parity.

snissn:

PR #1281, opened 2026-05-04
collections: prove replace root minimization

snissn:
## Summary

Adds a focused #1242 PR 2 replacement-update proof for the public `Replace` API:

- full replacement that changes only a non-indexed field changes the primary root but leaves persisted index-state and secondary roots unchanged;
- full replacement that changes only the `city` index updates primary/index-state/city roots while leaving the unrelated unique `email` root unchanged;
- verifies secondary index reads for old/new values after each replacement.

This is test-only and does not change update behavior.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionReplaceSkipsUnchangedSecondaryRoots' -count=1
```

### DIFF
--- a/TreeDB/collections/replace_minimization_test.go
+++ b/TreeDB/collections/replace_minimization_test.go
@@ -1,0 +1,138 @@
+package collections
+
+import (
+	"bytes"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


snissn:

PR #1283, opened 2026-05-04
collections: prove indexed flush root mismatch visibility

snissn:
## Summary

Adds a focused #1242 PR 1a root-base mismatch proof:

- one collection manager stages an indexed pending write;
- a second manager advances the same durable collection root;
- the first manager's async indexed publish preparation rejects the stale base root with `ErrConcurrentMutation`;
- pending primary and secondary reads for the staged write remain visible after the rejected prepare.

No production behavior changes.

## Tests

```sh
go test ./TreeDB/collections -run 'TestCollectionIndexedAsyncPrepareRejectsRootBaseMismatchAndPreservesPendingVisibility' -count=1
```

### DIFF
--- a/TreeDB/collections/indexed_flush_root_mismatch_test.go
+++ b/TreeDB/collections/indexed_flush_root_mismatch_test.go
@@ -1,0 +1,80 @@
+package collections
+
+import (
+	"bytes"
+	"errors"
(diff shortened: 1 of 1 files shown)


snissn:

PR #1285, opened 2026-05-04
collections: expose PR1B write-domain counters

snissn:
## Summary
- add PR1B collection write-domain counters for indexed flush units, overlay queued/active/mutable depth, root-delta plan roots/entries/bytes/tombstones, and primary-only no-index update publishes
- wire counters at existing indexed flush and direct/update-batch publish boundaries without changing publish behavior
- report the new counters from the Mongo gateway profile benchmark helper, including indexed_flush_units/batch and primary_root_delta_* metrics

## Tests
- go test ./TreeDB/collections -count=1
- go test ./cmd/mongo_gateway_bench -count=1
- go test ./cmd/internal/treedbstats -count=1

### DIFF
--- a/TreeDB/collections/api.go
+++ b/TreeDB/collections/api.go
@@ -1285,12 +1285,13 @@ func (domain *collectionWriteDomain) statsSnapshot() CollectionManagerStats {
 	domain.mu.RLock()
 	stats.PendingDocuments = domain.count
 	stats.PendingBytes = domain.bufferedBytes
-	stats.PendingRootRuns = bufferedIndexedRootRunCount(domain)
+	pendingRootRuns := bufferedIndexedRootRunCount(domain)
+	stats.PendingRootRuns = pendingRootRuns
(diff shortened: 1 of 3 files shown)


snissn:
Addressed the concrete Copilot metric/reporting comments in `4a796a691f`:

- `PrimaryOnlyBufferedCalls` now reports as `primary_only_buffered_calls` and `primary_only_buffered_calls/doc`.
- Profile metric tests now assert template-root, index-state-root, tombstone, and buffered primary-only metrics.
- Collection stats export test now asserts the missing root-delta and primary-only keys.
- `statsSnapshot()` now reuses the already computed pending root-run count for visible depth instead of scanning buffered runs twice.

Left the root-delta accounting implementation intentionally exact for PR1B; it still derives entry/key/value/tombstone counts from materialized ordered deltas. I included that tradeoff in the GPT-5 Pro packet as a specific code-review question.

snissn:
Superseded by aggregate PR #1330 so #1242 can be reviewed by phase instead of as many small shard PRs.

PR #1289, opened 2026-05-04
bench: include 32-writer scaling default

snissn:
#1242 / PR1B measurement cleanup. The required Mongo gateway writer sweep is writers `1/2/4/8/16/32`, but `scripts/mongo_gateway_scaling_bench.sh` still defaulted to `1/2/4/8/16`.

Changes:
- Add `32` to the default `WRITERS_LIST`.
- Update `--writers` help text to match.

Local checks:
- `bash -n scripts/mongo_gateway_scaling_bench.sh`
- `grep -n 'WRITERS_LIST=.*1 2 4 8 16 32\\|Default: "1 2 4 8 16 32"' scripts/mongo_gateway_scaling_bench.sh`

PR #1290, opened 2026-05-04
bench: allow writer-only scaling sweeps

snissn:
Stacked on #1289. This makes the dedicated #1242 writer sweep easier to run without spending time on reader-scaling cells.

Changes:
- Add `--no-reader-sweep`.
- Add `RUN_READER_SWEEP` environment override with the same bool parsing as other script booleans.
- Record the reader-sweep setting in stdout and the bundle README.

Local checks:
- `bash -n scripts/mongo_gateway_scaling_bench.sh`
- Tiny smoke: `scripts/mongo_gateway_scaling_bench.sh --out "$OUT" --docs 10 --indexes 0 --batch-size 5 --insert-producers 1 --writers "1" --no-reader-sweep --concurrent-writes 3 --concurrent-reads 1 --timeout 2m`; verified `matrix.tsv` contained only the writer row and no `readers_` row.


snissn:
Superseded by aggregate PR #1330 so #1242 can be reviewed by phase instead of as many small shard PRs.

PR #1292, opened 2026-05-04
collections: prove template-v1 template-only skips secondary roots

snissn:
## Summary

- add a PR2 regression proof for template-v1 direct-buffered updates where only the template shape changes
- assert primary/template roots are staged while unchanged unique and non-unique secondary roots are not staged
- additionally assert no pending unique-value reservation run is created for the unchanged unique field
- verify buffered reads and secondary lookups still see the expected logical document before flush, then flush succeeds

## Why

#1242 PR2 calls out template-v1 as a path where template-root work must stay separate from secondary-index and unique-reservation work. #1247 already covers the root-run separation side in a helper file; this PR adds the narrower pending-reservation assertion (`uniqueValueRuns == 0`) and flush boundary around the same direct-buffered template-only shape.

## Tests

```sh
go test ./TreeDB/collections -run '^TestCollectionUpdateBatchDirectBufferedTemplateV1TemplateOnlySkipsSecondaryRoots$' -count=1

go test ./TreeDB/collections -run 'TestCollectionUpdateBatchDirectBufferedTemplateV1(AccumulatesRootRuns|TemplateOnlySkipsSecondaryRoots)|TestTemplateV1UpdatePublishesNewTemplateShape' -count=1

```

### DIFF
--- a/TreeDB/collections/api_test.go
+++ b/TreeDB/collections/api_test.go
@@ -8494,6 +8494,113 @@ func TestCollectionUpdateBatchDirectBufferedTemplateV1AccumulatesRootRuns(t *tes
 	}
 }
 
+func TestCollectionUpdateBatchDirectBufferedTemplateV1TemplateOnlySkipsSecondaryRoots(t *testing.T) {
+	d, err := backenddb.Open(backenddb.Options{Dir: t.TempDir()})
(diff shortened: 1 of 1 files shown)


snissn:

PR #1295, opened 2026-05-04
collections: prove BSON replace id rejection

snissn:
## Summary

- add a PR2 replacement-path BSON `_id` mutation rejection proof for the public `Replace` API
- assert the rejected replacement does not advance commit sequence or change the primary root descriptor
- verify the original document remains readable and the mutated `_id` document is not created

## Why

#1242 PR2 calls out replacement update semantics and BSON `_id` mutation rejection. #1248 covers the direct `Update` callback path; this pins the public `Replace` API surface that wraps that path.

## Tests

```sh
go test ./TreeDB/collections -run '^TestCollectionReplaceBSONRejectsIDMutationBeforeRootWork$' -count=1

go test ./TreeDB/collections -run 'TestCollection(ReplaceBSONRejectsIDMutationBeforeRootWork|UpdateBSONRejectsIDMutationBeforeRootWork)|TestCollectionUpdateBatchBSONRejectsIDMutation' -count=1

```

### DIFF
--- a/TreeDB/collections/bson_replace_id_test.go
+++ b/TreeDB/collections/bson_replace_id_test.go
@@ -1,0 +1,91 @@
+package collections
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	backenddb "github.com/snissn/gomap/TreeDB/db"
+	"go.mongodb.org/mongo-driver/v2/bson"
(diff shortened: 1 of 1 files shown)


snissn:

PR #1298, opened 2026-05-04
collections: prove changed unique update drains pending

snissn:
## Summary

- add a PR1a guardrail test for changed secondary-unique direct updates while pending unique inserts exist
- cover both mutable pending indexed runs and queued indexed flush units
- assert the current safety model: pending writes are force-drained before durable unique preflight rejects the changed unique update

## Why

#1242 PR1a explicitly keeps changed-unique update admission out of pending layers until later semantic coalescing work. Current safety comes from flushing/draining pending writes before direct changed-unique preflight. This test pins that behavior and verifies the rejected update leaves the original document unchanged while the pending reservation becomes durable/visible.

## Tests

```sh
go test ./TreeDB/collections -run '^TestCollectionUpdateChangedUniqueFlushesPendingReservationsBeforeConflict$' -count=1

go test ./TreeDB/collections -run 'TestCollectionUpdateChangedUniqueFlushesPendingReservationsBeforeConflict|TestCollectionIndexedWriteMemtablesAsyncQueuedUnitsParticipateInUniqueChecks|TestCollectionUpdateBatchIfNoSecondaryUniqueIndexChangesDeclinesUniqueUpdates' -count=1

```

### DIFF
--- a/TreeDB/collections/changed_unique_pending_test.go
+++ b/TreeDB/collections/changed_unique_pending_test.go
@@ -1,0 +1,111 @@
+package collections
+
+import (
+	"bytes"
+	"errors"
+	"testing"
+
+	backenddb "github.com/snissn/gomap/TreeDB/db"
+)
+
+func TestCollectionUpdateChangedUniqueFlushesPendingReservationsBeforeConflict(t *testing.T) {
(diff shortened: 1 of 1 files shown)


snissn:

PR #1300, opened 2026-05-04
collections: prove int64 update batch minimization

snissn:
## Summary

- add a PR2 UpdateBatch minimization proof for an unchanged large JSON int64 index value
- use `9007199254740993` to pin exact int64 extraction rather than float-style rounding
- verify only the primary root moves while the persisted index-state and `score` secondary roots stay stable

## Why

#1242 PR2 calls out same-effective-value coverage, including numeric exactness according to `IndexValueType`. Existing planner/index tests cover int64 extraction and range behavior; this adds a collection update-root minimization proof for unchanged effective int64 values.

## Tests

```sh
go test ./TreeDB/collections -run '^TestCollectionUpdateBatchSkipsUnchangedInt64IndexRoot$' -count=1

go test ./TreeDB/collections -run 'TestCollectionUpdateBatchSkipsUnchangedInt64IndexRoot|TestCollectionIndexInt64JSON|TestCollectionFindByIndexRangeInt64|TestCollectionBufferedRangeFindIncludesPendingInt64Updates' -count=1

```

### DIFF
--- a/TreeDB/collections/update_int64_minimization_test.go
+++ b/TreeDB/collections/update_int64_minimization_test.go
@@ -1,0 +1,105 @@
+package collections
+
+import (
+	"bytes"
+	"testing"
+
+	backenddb "github.com/snissn/gomap/TreeDB/db"
+)
+
+func TestCollectionUpdateBatchSkipsUnchangedInt64IndexRoot(t *testing.T) {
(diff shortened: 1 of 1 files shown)


snissn:

PR #1311, opened 2026-05-04
collections: prove template-v1 direct double minimization

snissn:
## Summary

- adds a PR2 direct-update proof for template-v1 double index minimization
- verifies a same-shape replacement that keeps the same double indexed value moves only the primary root
- verifies changing the double indexed value moves the secondary root while preserving the template root for the same shape

## Why

#1242 PR2 asks for numeric exactness proof across formats and paths. This complements the template-v1 int64 rows with the `IndexValueDouble` extraction/encoding path.

## Tests

```sh
go test ./TreeDB/collections -run '^TestCollectionDirectUpdateTemplateV1DoubleSameEffectiveSkipsSecondaryRoot$' -count=1

go test ./TreeDB/collections -run 'Test(CollectionDirectUpdateTemplateV1DoubleSameEffectiveSkipsSecondaryRoot|CollectionIndexesCanonicalExtendedJSONNumbers|CollectionTemplateV1UpdateSeparatesTemplateAndSecondaryRootWork|TemplateV1MultiKeyIndex)$' -count=1

```

### DIFF
--- a/TreeDB/collections/template_v1_direct_double_minimization_test.go
+++ b/TreeDB/collections/template_v1_direct_double_minimization_test.go
@@ -1,0 +1,152 @@
+package collections
+
+import (
+	"sort"
+	"testing"
+
+	backenddb "github.com/snissn/gomap/TreeDB/db"
+)
+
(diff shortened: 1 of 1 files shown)


snissn:

PR #1412, opened 2026-05-07
Collections: batch indexed range document reads

snissn:
## Summary

Adds a native collection helper for indexed range reads that returns primary documents directly:

- `Collection.FindDocumentsByIndexRange` scans the secondary index and fetches matching primary documents under one persisted snapshot/catalog;
- same-manager buffered documents remain visible by checking the collection write-domain before the persisted primary root;
- direct Mongo parity benchmarks now use the native helper for indexed range reads instead of `FindByIndexRange` plus one `Collection.Get` call per returned id.

This is stacked on #1405.

## Local validation

```text
GOWORK=off go test -count=1 ./TreeDB/collections ./cmd/mongo_gateway_bench
git diff --check
```

## Direct range-read smoke

Compared against #1405 head with settled reads, `age_1` index, `documents=20000`, `range_reads=10000`, all non-range phases disabled.

| Format | #1405 ops/sec | PR ops/sec | #1405 ns/op | PR ns/op | Throughput delta |
| JSON | 16,541 | 16,698 | 60,403 | 59,827 | +0.9% |
| TemplateV1 | 11,041 | 11,862 | 90,512 | 84,249 | +7.4% |
| BSON | 94,773 | 102,233 | 10,500 | 9,735 | +7.9% |

Artifacts:

```text
/tmp/mongo_range_pr2_direct_json.json
/tmp/mongo_range_pr3_direct_json.json
/tmp/mongo_range_pr2_direct_template-v1.json
/tmp/mongo_range_pr3_direct_template-v1.json
/tmp/mongo_range_pr2_direct_bson.json
/tmp/mongo_range_pr3_direct_bson.json
```

### DIFF
--- a/TreeDB/collections/api.go
+++ b/TreeDB/collections/api.go
@@ -11457,17 +11457,32 @@ func (c *Collection) FindDocumentsByIndexRange(indexName string, opts IndexRange
 		// buffered primary documents come from one write-domain view. Releasing
 		// here would require copying the pending primary view up front, which is
 		// too much work for the common small-limit range probe.
-		defer func() {
-			if domainLocked {
-				domain.mu.RUnlock()
-			}
-		}()
 	}
-	snap := c.db.AcquireSnapshot()
+	var snap *backenddb.Snapshot
+	var bufferedTable memtable.Table
+	var bufferedIt iterator.UnsafeIterator
+	var persistedIt iterator.UnsafeIterator
+	defer func() {
+		if domainLocked {
+			domain.mu.RUnlock()
+		}
+		if persistedIt != nil {
+			_ = persistedIt.Close()
+		}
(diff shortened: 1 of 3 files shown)


snissn:
Review follow-up on write-domain lock scope:

I kept the read lock through the small-limit range document scan intentionally. This path has to keep buffered secondary IDs and buffered primary documents tied to one write-domain view; releasing after materializing only the secondary table can pair an older/newer buffered primary view with the secondary IDs. A correct unlocked version would need to snapshot/copy the pending primary view up front, which is a larger design change and would risk the limit-10 perf work in this stack. I added an inline comment documenting that tradeoff so we can revisit with a dedicated snapshot API if broad range scans show write blocking in profiles.

PR #1414, opened 2026-05-07
Mongo gateway: reduce range read response copies

snissn:
## Summary

Stacked on #1412. This trims the Mongo gateway indexed range-read response path:

- use `Collection.FindDocumentsByIndexRange` in the gateway indexed range materialization path instead of scanning ids and then issuing separate primary `Get` calls;
- build raw cursor BSON in one buffer for raw first-batch responses;
- for OP_MSG `find` responses with raw BSON documents, build the OP_MSG response directly instead of first building a command document and then copying it into the OP_MSG body.

This intentionally does **not** add a gateway collection lookup cache.

## Validation

```bash
go test -count=1 ./TreeDB/mongo_gateway ./cmd/mongo_gateway_bench ./TreeDB/collections
```

## Range-read benchmark

Workload: TreeDB target, BSON collection storage, settled reads, 20k loaded docs, age indexed range query with `limit=10`, 50k single-flight range operations, `-treedb-maintenance none`.

Artifacts:

- #1412 baseline: `/Users/michaelseiler/dev/snissn/benchdata/mongo_4levels_20260506_223314`
- current PR: `/Users/michaelseiler/dev/snissn/benchdata/mongo_4levels_opt3_20260506_224606`

| Mode | #1412 ops/s | #1412 ns/op | Current ops/s | Current ns/op | Delta |
|---|---:|---:|---:|---:|---:|
| driver-command-raw | 21,735 | 46,010 | 22,330 | 44,783 | +2.7% |
| raw-wire-tcp | 21,535 | 46,436 | 24,721 | 40,451 | +14.8% |
| raw-wire | 70,920 | 14,100 | 77,790 | 12,855 | +9.7% |
| direct | 105,103 | 9,514 | 101,310 | 9,871 | -3.6% |

The direct row should be treated as a noise/control row for this PR; the code path changed here is gateway-only.

## Fanout sweep

Workload: same setup, but `-concurrent-range-reads 80000 -concurrent-range-reader-sweep 1,2,4,8,16,32,64`.

Artifacts: `/Users/michaelseiler/dev/snissn/benchdata/mongo_range_concurrency_opt3_20260506_224710`

| Mode | R1 ops/s | R1 ns/op | Best readers | Best ops/s | Best ns/op |
|---|---:|---:|---:|---:|---:|
| driver-command-raw | 22,050 | 45,352 | 16 | 86,677 | 11,537 |
| raw-wire-tcp | 24,979 | 40,033 | 32 | 89,348 | 11,192 |
[...]

PR #1415, opened 2026-05-07
Collections: borrow range scan ids for document reads

snissn:
## Summary

Stacked on #1414. This trims the native indexed range document scan path:

- keep existing public `ScanIndexRange`/`FindByIndexRange` owned-id behavior;
- let `FindDocumentsByIndexRange` use an internal borrowed-id scan so the id is not cloned once by the scan and then cloned again into the returned `DocumentRecord`;
- skip the dedupe map/string allocation entirely when there is no buffered index iterator, which is the settled persisted-read case used by the range benchmarks.

The returned `DocumentRecord` values from `FindDocumentsByIndexRange` remain owned just as before.

## Validation

```bash
go test -count=1 ./TreeDB/collections ./TreeDB/mongo_gateway ./cmd/mongo_gateway_bench
```

## Range-read benchmark

Workload: TreeDB target, BSON collection storage, settled reads, 20k loaded docs, age indexed range query with `limit=10`, 50k single-flight range operations, `-treedb-maintenance none`.

Artifacts:

- #1414 baseline: `/Users/michaelseiler/dev/snissn/benchdata/mongo_4levels_opt3_20260506_224606`
- current PR: `/Users/michaelseiler/dev/snissn/benchdata/mongo_4levels_opt4_20260506_225235`

| Mode | #1414 ops/s | #1414 ns/op | Current ops/s | Current ns/op | Delta |
| driver-command-raw | 22,330 | 44,783 | 22,432 | 44,578 | +0.5% |
| raw-wire-tcp | 24,721 | 40,451 | 24,826 | 40,281 | +0.4% |
| raw-wire | 77,790 | 12,855 | 81,962 | 12,201 | +5.4% |
| direct | 101,310 | 9,871 | 107,954 | 9,263 | +6.6% |

This mostly helps the lower two layers where native range scan work is visible. It barely moves driver/TCP single-flight, which is expected because those remain dominated by socket/request-response costs.

## Profile notes

Before this change, the direct range profile showed `scanMergedCollectionIndexIDs` cloning ids before the callback, then `FindDocumentsByIndexRange` cloning the same ids again for the returned owned records. After this change, the settled persisted-only path avoids the scan clone and avoids the dedupe map/string allocation.

The remaining large native allocation is the owned document clone in `FindDocumentsByIndexRange`. The next meaningful direct-path optimization is a borrowed document scan/cursor API for callers that consume documents during the scan and do not need an owned result slice.

### DIFF
--- a/TreeDB/collections/api.go
+++ b/TreeDB/collections/api.go
@@ -11851,12 +11851,12 @@ func scanMergedCollectionIndexIDsWithOptions(bufferedIt, persistedIt iterator.Un
 				persistedIt.Next()
 				continue
 			}
+			if maxResults > 0 && emitted >= maxResults {
+				return true, nil
+			}
 			id, err := indexKeyDocumentID(valueType, persistedKey)
 			if err != nil {
 				return false, err
-			}
-			if maxResults > 0 && emitted >= maxResults {
-				return true, nil
 			}
 			if cloneID {
 				id = bytes.Clone(id)
@@ -11870,18 +11870,23 @@ func scanMergedCollectionIndexIDsWithOptions(bufferedIt, persistedIt iterator.Un
 		}
 		return false, collectionIndexIteratorError(persistedIt)
 	}
-	seen := make(map[string]struct{})
+	var seen map[string]struct{}
+	if dedupeIDs {
+		seen = make(map[string]struct{})
+	}
 	emitted := 0
 	emit := func(key []byte) (bool, bool, error) {
 		id, err := indexKeyDocumentID(valueType, key)
 		if err != nil {
 			return false, false, err
 		}
-		idKey := string(id)
-		if _, ok := seen[idKey]; ok {
-			return true, false, nil
+		if dedupeIDs {
+			idKey := string(id)
+			if _, ok := seen[idKey]; ok {
+				return true, false, nil
+			}
+			seen[idKey] = struct{}{}
 		}
-		seen[idKey] = struct{}{}
 		if maxResults > 0 && emitted >= maxResults {
 			return false, true, nil
 		}
--- a/TreeDB/collections/api.go
+++ b/TreeDB/collections/api.go
@@ -11849,17 +11849,20 @@ func encodedDoubleComponentIsNaN(encoded []byte) bool {
 	return len(encoded) == 1 && encoded[0] == 0x00
 }
 
-func scanMergedCollectionIndexIDs(bufferedIt, persistedIt iterator.UnsafeIterator, valueType IndexValueType, maxResults int, dedupeDocumentIDs bool, fn func([]byte) (bool, error)) (bool, error) {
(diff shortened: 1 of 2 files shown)


snissn:
### Range-read fanout sweep

Artifact: `/Users/michaelseiler/dev/snissn/benchdata/mongo_range_concurrency_opt4_20260506_225733`

Config: TreeDB target, BSON collection storage, settled reads, `20k` docs, `age_1` indexed range `limit: 10`, `80k` concurrent range reads per fanout row, maintenance disabled, prebuilt load docs. `ns/op` below is wall-equivalent throughput cost: `1e9 / ops/sec`.

| Mode | Best readers | Best ops/sec | Best ns/op |
|---|---:|---:|---:|
| driver-command-raw | 32 | 67,664 | 14,780 |
| raw-wire-tcp | 32 | 66,701 | 14,993 |
| raw-wire | 4 | 147,323 | 6,788 |
| direct | 4 | 208,668 | 4,793 |

| Mode | Readers | ops/sec | ns/op |
|---|---:|---:|---:|
| driver-command-raw | 1 | 23,113 | 43,267 |
| driver-command-raw | 2 | 40,690 | 24,576 |
| driver-command-raw | 4 | 48,526 | 20,608 |
| driver-command-raw | 8 | 61,970 | 16,137 |
| driver-command-raw | 16 | 65,766 | 15,205 |
| driver-command-raw | 32 | 67,664 | 14,780 |
| driver-command-raw | 64 | 58,568 | 17,074 |
| raw-wire-tcp | 1 | 25,122 | 39,808 |
| raw-wire-tcp | 2 | 35,986 | 27,790 |
| raw-wire-tcp | 4 | 49,253 | 20,304 |
| raw-wire-tcp | 8 | 59,281 | 16,869 |
| raw-wire-tcp | 16 | 65,823 | 15,192 |
| raw-wire-tcp | 32 | 66,701 | 14,993 |
| raw-wire-tcp | 64 | 66,431 | 15,053 |
| raw-wire | 1 | 69,116 | 14,469 |
| raw-wire | 2 | 122,321 | 8,176 |
| raw-wire | 4 | 147,323 | 6,788 |
| raw-wire | 8 | 116,032 | 8,619 |
| raw-wire | 16 | 113,830 | 8,785 |
| raw-wire | 32 | 104,505 | 9,569 |
| raw-wire | 64 | 101,332 | 9,869 |
| direct | 1 | 98,145 | 10,189 |
| direct | 2 | 163,731 | 6,107 |
| direct | 4 | 208,668 | 4,793 |
| direct | 8 | 172,807 | 5,787 |
| direct | 16 | 162,379 | 6,159 |
| direct | 32 | 151,054 | 6,620 |
| direct | 64 | 107,018 | 9,344 |

Skeptical read: this fanout run is lower than the earlier parent-head fanout sweep, while the single-flight four-level run on this PR still showed the intended direct/raw-wire improvement. I would not claim a fanout throughput win from this PR based on this one run; it mainly confirms the optimization does not change the scaling shape and that high-concurrency results are noisy enough to require repeated runs before making a concurrency claim.

PR #1416, opened 2026-05-07
Collections: add borrowed indexed range document scan

snissn:
## Summary

Adds `Collection.ScanBorrowedDocumentsByIndexRange`, a callback API for indexed range reads that borrows the matched primary document slices for the duration of the callback.

The existing `FindDocumentsByIndexRange` API still returns owned `DocumentRecord` slices and preserves its clone/ownership contract by delegating through the borrowed scanner and cloning at the boundary.

The direct Mongo-gateway parity benchmark now uses the borrowed scan for indexed direct range reads, since it consumes each document immediately and does not need to retain an owned result slice.

## Why

On #1415, the remaining direct/native indexed range allocation was the owned document clone inside `FindDocumentsByIndexRange`. That clone is correct for the owned API, but it is unnecessary for immediate consumers such as the direct benchmark and future streaming response builders.

## Validation

Local tests:

```bash
go test -count=1 ./TreeDB/collections ./cmd/mongo_gateway_bench ./TreeDB/mongo_gateway
```

Result: passed.

## Benchmark

Config: TreeDB target, direct client mode, BSON collection storage, settled reads, `20k` docs, `age_1` indexed range `limit: 10`, `50k` range reads, maintenance disabled, prebuilt load docs.

| Branch | ops/sec | ns/op |
|---|---:|---:|
| #1414 parent | 101,310 | 9,871 |
| #1415 borrowed id scan | 107,954 | 9,263 |
| this PR borrowed document scan | 120,588 | 8,293 |

Artifact: `/Users/michaelseiler/dev/snissn/benchdata/mongo_direct_borrowed_docs_20260506_230223`

Profiled validation run:

| Branch | ops/sec | ns/op |
|---|---:|---:|
| this PR profiled direct run | 116,759 | 8,565 |

Artifact: `/Users/michaelseiler/dev/snissn/benchdata/mongo_direct_borrowed_docs_profile_20260506_230249`

The range-document clone site from `FindDocumentsByIndexRange` is no longer on the direct indexed range path. Remaining allocation profile entries are mostly setup/load costs captured by the broad harness profile, not isolated direct range inner-loop allocations.

PR #1417, opened 2026-05-07
mongo bench: trust stored BSON in direct reads

snissn:
## Summary

Updates the direct TreeDB Mongo-gateway benchmark path to trust stored BSON bytes instead of validating them on every read.

This matches the production gateway hot path, where stored BSON is validated when accepted or built and `storedDocumentToBSON` returns the stored bytes directly for BSON document collections.

## Why

After #1416 removes the owned document clone from direct indexed range reads, the direct profile still showed benchmark-side `bson.Raw.Validate` as a visible cost. That validation is not part of the gateway read path and made the direct/native baseline look slower than the server path it is meant to isolate.

## Validation

```bash
go test -count=1 ./cmd/mongo_gateway_bench
```

Result: passed.

## Benchmark

Config: TreeDB target, direct client mode, BSON collection storage, settled reads, `20k` docs, `age_1` indexed range `limit: 10`, `50k` range reads, maintenance disabled, prebuilt load docs.

| Branch | ops/sec | ns/op |
|---|---:|---:|
| #1415 borrowed id scan | 107,954 | 9,263 |
| #1416 borrowed document scan | 120,588 | 8,293 |
| this PR trust stored BSON | 131,393 | 7,611 |

Artifact: `/Users/michaelseiler/dev/snissn/benchdata/mongo_direct_trust_bson_20260506_230649`

This is a measurement cleanup for the direct/native benchmark layer. It should not be presented as a production gateway throughput improvement by itself.

### DIFF
--- a/cmd/mongo_gateway_bench/main.go
+++ b/cmd/mongo_gateway_bench/main.go
@@ -1873,11 +1873,9 @@ func directStoredDocumentToBSON(collection *collections.Collection, materializer
 		return nil, errors.New("direct TreeDB benchmark requires a document materializer")
 	}
 	if materializer.DocumentFormat() == collections.DocumentFormatBSON {
-		raw := bson.Raw(stored)
-		if err := raw.Validate(); err != nil {
-			return nil, err
-		}
-		return raw, nil
+		// Match the gateway hot read path: stored BSON was validated on write,
+		// so direct read timing should not include repeated full validation.
+		return bson.Raw(stored), nil
 	}
 	materialized, err := materializer.StoredDocumentJSON(stored)
 	if err != nil {


snissn:
## Local direct range fanout benchmark

Artifact: `/Users/michaelseiler/dev/snissn/benchdata/mongo_direct_trust_bson_fanout_20260506_232409`

Workload:

```text
mongo_gateway_bench
  -target treedb
  -client-mode direct
  -documents 20000
  -batch-size 1000
  -secondary-indexes 2
  -range-index
  -treedb-document-format bson
  -treedb-maintenance none
  -treedb-read-state settled
  -prebuild-documents
  -concurrent-range-reads 100000
  -concurrent-range-reader-sweep 1,2,4,8,16,32,64
```

Results for `age_range_indexed_limit_10`:

| Readers | ops/sec | ns/op |
|---:|---:|---:|
| 1 | 138,645 | 7,212 |
| 2 | 231,756 | 4,314 |
| 4 | 332,099 | 3,011 |
| 8 | 286,462 | 3,490 |
| 16 | 270,649 | 3,694 |
| 32 | 255,913 | 3,907 |
| 64 | 222,837 | 4,487 |

This is still benchmark harness parity work, not a production gateway optimization. It makes the direct BSON benchmark stop paying `bson.Raw.Validate()` on stored docs, and direct settled range reads now peak around `332k ops/sec` / `3.0us/op` at R4 in this local run.

PR #1432, opened 2026-05-08
Fix value-log rewrite grouping and byte stats

### LINKED ISSUE #1430: TreeDB maintenance UX: value_vlog/leaf_vlog rewrite paths are confusing and offline rewrite can leave missing leaf_vlog refs
## Summary

While building a JSONBench raw-TreeDB storage experiment, I hit confusing and partly surprising TreeDB maintenance behavior around `value_vlog`, `leaf_vlog`, `vlog-gc`, online rewrite, offline `vlog-rewrite`, leaf-generation pack/GC, and final disk-size reporting.

This is at least a UX/documentation/API issue, and one path looks like a correctness bug: offline `ValueLogRewriteOffline` followed by leaf-generation maintenance can leave the DB referencing a missing `leaf_vlog` file.

Tested against gomap commit `7a3d4701691db8f8ddafe3885318ed363ca048f0`.

## Why this is confusing

The CLI exposes these commands:

```text
vlog-gc         Delete unreferenced value-log segments
vlog-rewrite    Rewrite value-log segments and shrink via swap
leafgen-plan    Print explicit leaf-generation pack plan
leafgen-pack    Pack sealed leaf generations by id
leafgen-gc      Delete fully unreachable sealed leaf generations
```

For a caller trying to produce a fully compacted disk-size number, it is not clear whether to use:

- public cached `DB.ValueLogRewriteOnline` + `DB.ValueLogGC`
- backend-direct `ValueLogRewriteOnline` + `ValueLogGC`
- offline `treedb.ValueLogRewriteOffline` / `treemap vlog-rewrite`
- `leafgen-pack` / `leafgen-gc`
- online or offline index vacuum

The names also overlap: `vlog-rewrite` sounds value-only, but the offline implementation appears to also handle outer-leaf/leaf-log rewrite when `IndexOuterLeavesInValueLog` is enabled. `vlog-gc` sounds like the compaction step, but it only deletes already-unreferenced value-log segments and does not compact live values.

## Observed behavior 1: public online rewrite + GC can grow disk and GC does not reclaim

A 1M raw JSON load into raw TreeDB with `ProfileFast`, `ValueLog.PointerThreshold=1`, and `ForcePointers=true` produced:

```text
after_load_checkpoint:   224.70 MiB
after_value_log_rewrite: 679.70 MiB
after_value_log_gc:      679.70 MiB
```

The GC stats showed the old large segment as active, so it was not eligible for deletion. That may be correct for the public cached online API, but it is surprising when the user asked for a post-load compacted disk footprint.

## Observed behavior 2: backend-direct online rewrite reports tiny live bytes but large filesystem bytes

Using backend-direct maintenance after close, with value-log rewrite, value-log GC, leaf-generation pack/GC, and offline index vacuum, the 1M raw JSON run produced:

```text
raw JSON bytes:                    457.55 MiB
after_load_checkpoint:             208.47 MiB
after_value_log_rewrite:           468.68 MiB
after_value_log_gc:                468.68 MiB
after_leaf_generation_pack:        468.68 MiB
after_leaf_generation_gc:          459.11 MiB
after_index_vacuum_offline:        479.11 MiB
ClickHouse JSONBench 1M baseline:   97.07 MiB
DuckDB JSONBench 1M baseline:      462.75 MiB
```

But the value-log rewrite stats reported much smaller live/referenced bytes:

```json
{
  "SegmentsBefore": 6,
  "SegmentsAfter": 4,
  "BytesBefore": 204304588,
  "BytesAfter": 28291936,
  "RecordsCopied": 1000000,
  "ValueRecordsCopied": 1000000,
  "ValueBytesCopied": 479778277
}
```

The filesystem contained four preallocated value-log files (`128M`, `128M`, `128M`, `60M`). So the reported compacted disk footprint is dominated by segment allocation policy, while rewrite stats make the data look closer to 28 MiB. This makes benchmark-size reporting hard to reason about.

With a 32 MiB online rewrite segment cap, the result still ended around 479 MiB because the rewrite created many 32 MiB files; stats changed to `BytesAfter=65492255`, still far below file-size disk usage.

## Observed behavior 3: offline `ValueLogRewriteOffline` then leafgen pack can fail with missing leaf_vlog file

A small raw TreeDB smoke using `ProfileFast`, forced pointers, checkpoint, close, then `treedb.ValueLogRewriteOffline`, then `OpenBackendWithCachedLeafLog` + `LeafGenerationPackRunOnce` failed with:

```text
pack leaf generations: open /tmp/jsonbench_raw_treedb_smoke3/db/maindb/leaf_vlog/value-l255-000001.log: no such file or directory
```

After the failure the directory contained `leaf_vlog/value-l255-000002.log` and `value-l255-000003.log`, but not `value-l255-000001.log`. It also contained many zero-byte `maindb/value_vlog/value-l<N>-000001.log` lane files, apparently created during the attempted backend/cached-leaf-log reopen.

This looks like offline rewrite may leave manifest/index state referring to a removed old leaf-log file, or the backend reopen/leafgen maintenance path is using stale leaf-log metadata.

## Standalone reproduction shape

This is the minimal shape I expect to reproduce without this machine:

```go
package main

import (
    "context"
    "encoding/binary"
    "fmt"
    "os"

    treedb "github.com/snissn/gomap/TreeDB"
    backenddb "github.com/snissn/gomap/TreeDB/db"
)

func main() {
    dir := os.Args[1]
    _ = os.RemoveAll(dir)

    opts := treedb.OptionsFor(treedb.ProfileFast, dir)
    opts.ValueLog.PointerThreshold = 1
    opts.ValueLog.ForcePointers = true

    db, err := treedb.Open(opts)
    if err != nil { panic(err) }
    for i := 0; i < 6; i++ {
        key := make([]byte, 8)
        binary.BigEndian.PutUint64(key, uint64(i+1))
        val := []byte(fmt.Sprintf(`{"id":%d,"payload":"abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyz"}`, i))
        if err := db.Set(key, val); err != nil { panic(err) }
    }
    if err := db.Checkpoint(); err != nil { panic(err) }
    if err := db.Close(); err != nil { panic(err) }

    stats, err := treedb.ValueLogRewriteOffline(opts)
    fmt.Printf("offline rewrite stats: %+v err=%v\n", stats, err)
    if err != nil { panic(err) }

    backend, cleanup, err := treedb.OpenBackendWithCachedLeafLog(opts)
    if err != nil { panic(err) }
    defer cleanup()

    pack, err := backend.LeafGenerationPackRunOnce(context.Background(), backenddb.LeafGenerationPackFromPlanOptions{Force: true, Sync: true})
    fmt.Printf("leafgen pack stats: %+v err=%v\n", pack, err)
    if err != nil { panic(err) }
}
```

Expected: the offline rewrite path should leave the DB readable and compatible with leaf-generation maintenance, or the documented maintenance command sequence should say not to run leafgen after offline rewrite.

Actual: leaf-generation maintenance can fail trying to open an old missing `leaf_vlog` segment.

## Expected UX

We need one documented command/API for "produce final compacted TreeDB disk size" that states exactly which storage domains it covers:

- `value_vlog` user values
- `leaf_vlog` outer leaf pages
- `index.db`
- side stores such as `dictdb` / `templatedb`

It should also make clear whether reported bytes are logical live bytes, value-log file sizes, allocated filesystem bytes, or benchmark-comparable total DB directory size.

snissn:
## Summary

Fixes the value-log rewrite regression documented in #1430 where rewrite could turn JSON-like pointer values into weak single-record Snappy frames and report misleading post-rewrite byte totals.

Changes:
- Batch dictID=0 rewrite values into grouped block-compressed frames, returning predicted grouped pointers and flushing at existing rewrite boundaries.
- Decode and regroup compressed k=1 source frames during iterator/offline rewrite instead of preserving weak single-record frames.
- Use actual file stat sizes for value-log rewrite/GC byte accounting so `BytesAfter`/GC totals match on-disk files after registered segments grow.
- Add regression tests for online grouping, offline grouping of compressed source frames, and actual byte reporting after segment growth.

## Evidence

Tests run locally:

```sh
go test ./TreeDB/db -count=1
go test ./TreeDB ./TreeDB/internal/valuelog -count=1
```

JSONBench raw TreeDB 1M confirmation against this branch:

```text
after_load_checkpoint:      234.37 MiB
after_value_log_rewrite:    235.61 MiB
after_leaf_generation_gc:   225.86 MiB
after_index_vacuum_offline: 245.86 MiB
value_vlog du:              212M
rewrite stats BytesAfter:   222,529,026 bytes
rewritten frames:           avg k ~= 128, payload ratio ~= 0.44
```

Before this fix, the same saved raw harness ended around 479 MiB total with value_vlog around 444 MiB and rewrite stats incorrectly reporting only 28.29 MB.

Fixes #1430.

### DIFF
--- a/TreeDB/db/vlog_gc.go
+++ b/TreeDB/db/vlog_gc.go
@@ -565,5 +565,10 @@ func fileSize(f *valuelog.File) int64 {
 	if f == nil {
 		return 0
 	}
+	if f.Path != "" {
+		if info, err := os.Stat(f.Path); err == nil {
+			return info.Size()
+		}
+	}
 	return f.SizeBestEffort()
 }
--- a/TreeDB/db/vlog_rewrite.go
+++ b/TreeDB/db/vlog_rewrite.go
@@ -38,6 +38,8 @@ const defaultValueLogRewriteSegmentBytes = 128 << 20
 
 const rewriteDictMinPayloadBytes = 32 << 10
 const rewriteDictBatchMaxK = 64
+const rewriteBlockBatchMaxK = valuelog.MaxFrameK
+const rewriteBlockBatchMaxRawBytes = 4 << 20
 const rewriteReadScratchMaxCap = 1 << 20 // 1MiB cap to avoid retaining oversized decode buffers
 const rewriteKeyArenaMaxCap = 1 << 20    // 1MiB cap to avoid retaining oversized key arenas
 
@@ -3052,7 +3054,7 @@ func ValueLogRewriteOffline(opts Options) (ValueLogRewriteStats, error) {
 		// Pointer mappings returned while grouped dict batches are still pending
 		// rely on batch-flush offset stability; flush here so record-count stats
 		// and subsequent tree builds observe committed batches.
-		if err := writer.flushPendingDictBatch(); err != nil {
+		if err := writer.flushPendingBatches(); err != nil {
 			return 0, err
 		}
 		stats.RecordsCopied = writer.records
@@ -3312,6 +3314,13 @@ type rewriteWriter struct {
 	createdIDs                []uint32
 	createdSegments           []rewriteCreatedSegment
 
+	pendingBlockStart   int64
+	pendingBlockRaw     int
+	pendingBlockArena   []byte
+	pendingBlockRecords []valuelog.Record
+	pendingBlockPtrs    []page.ValuePtr
+	pendingBlockDst     []page.ValuePtr
+
 	pendingDictID      uint64
 	pendingDict        []byte
 	pendingDictStart   int64
@@ -3366,6 +3375,25 @@ func (w *rewriteWriter) hasPendingDictBatch() bool {
 	return w != nil && len(w.pendingDictRecords) > 0
 }
 
+func (w *rewriteWriter) hasPendingBlockBatch() bool {
+	return w != nil && len(w.pendingBlockRecords) > 0
+}
+
+func (w *rewriteWriter) resetPendingBlockBatch() {
+	if w == nil {
+		return
+	}
+	w.pendingBlockStart = 0
+	w.pendingBlockRaw = 0
+	if cap(w.pendingBlockArena) > rewriteBlockBatchMaxRawBytes*2 {
+		w.pendingBlockArena = nil
+	} else {
+		w.pendingBlockArena = w.pendingBlockArena[:0]
+	}
+	w.pendingBlockRecords = w.pendingBlockRecords[:0]
+	w.pendingBlockPtrs = w.pendingBlockPtrs[:0]
+}
+
 func (w *rewriteWriter) resetPendingDictBatch() {
 	if w == nil {
 		return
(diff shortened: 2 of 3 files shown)


snissn:
@codex review

snissn:
@copilot review

PR #1454, opened 2026-05-09
Reclaim cached rewrite source segments

snissn:
## Objective

Ensure public cached value-log rewrite and compact-storage maintenance reclaim source segments after backend rewrite has proven those segments are unreferenced.

## Context

The JSONBench raw TreeDB fixture showed dictionary-compressed rewrite output plus retained original ingest segments. The original segments were still classified as active lane heads, so GC kept them even though rewrite reported them unreferenced.

## Non-goals

- No change to value-log compression codecs.
- No collection WAL work.
- No benchmark report changes in this PR.

## Design

- Add cached `ReclaimObservedValueLogSources`, which deduplicates backend rewrite source IDs, forces a cached checkpoint rotation, then runs observed-source GC for the proven-unreferenced IDs.
- Add backend `ObservedSourceReclaimActive` so observed-only GC can reclaim an otherwise-active segment only when the caller explicitly proves it is safe.
- Wire public `ValueLogRewriteOnline` and `CompactStorage` to invoke the cached reclaim hook when backend rewrite returns `SourceFileIDsUnreferenced`.

## Scope

- Includes: `TreeDB/caching.DB.ReclaimObservedValueLogSources`, backend `ValueLogGCOptions`, public `ValueLogRewriteOnline`, public `CompactStorage`, regression tests.
- Excludes: compression format changes, collection storage, benchmark artifact regeneration.

## Correctness

- Tests run:
  - `go test ./TreeDB -run TestCachedValueLogRewriteOnlineReclaimsObservedActiveSources -count=1`
  - `go test ./TreeDB/db -run TestValueLogGC_ObservedSourceReclaimActiveRequiresExplicitOption -count=1`
  - `go test ./TreeDB ./TreeDB/db`
- Corruption/edge cases covered: default observed-source GC still preserves active segments; active observed-source reclaim requires the explicit option.
- Concurrency/race coverage: cached public rewrite test fences through checkpoint before active-source reclaim and validates all rewritten values remain readable.

## Performance

- Benchmarks run: none.
- Results table: not applicable.
- Regressions: manual cached rewrite/compact may do one extra checkpoint and observed-source GC pass when rewrite returns unreferenced source IDs; this is the intended cleanup path.

## Rollout / Toggle Plan

- Default setting: enabled only for public cached rewrite/compact after backend rewrite reports unreferenced source IDs.
- How to disable quickly: revert the public wrapper calls to `ReclaimObservedValueLogSources`.

## Follow-ups

- Rerun the JSONBench raw TreeDB fixture after this PR lands to update the storage report.

---

### Scope
- [x] Single, clear objective for this PR (not a bundle)
- [x] Explicit include list (files/symbols touched): public rewrite/compact wrappers, cached reclaim hook, backend observed-source GC option, tests
- [x] Explicit exclude list (files/symbols NOT touched): compression codecs, collection WAL, benchmark report
- [x] No unwanted feature reintroduction (e.g. async slab writer, hard zones, ValueIndex) unless explicitly stated in the PR objective

### Safety / Correctness
- [x] Clear written-vs-durable semantics for any `*Sync` path touched
- [x] No new mmap usage on mutable/truncating files
- [x] All new length fields are capped before allocation (OOM-safe)
- [x] CRC/checksum behavior is fail-closed (clean error, no panic)
- [x] Any concurrency change has a defined state machine and invariants

### Tests
- [x] Added deterministic regression tests for the targeted failure mode
- [x] Tests avoid sleeps where possible (use latches/hooks)
- [ ] Added corruption/edge-case tests (truncation, bad headers, invalid lengths) when format/parsing changes
- [x] Ran locally:
  - [ ] `go test ./... -count=1`
  - [x] Any targeted packages/benchmarks: `go test ./TreeDB ./TreeDB/db`

### Benchmarks / Measurements
- [ ] Added or updated a benchmark relevant to the change
- [ ] Reported results in PR description (before/after, command, machine)
- [ ] Included “incompressible” (worst-case) results if compression is involved

### Docs
- [ ] Updated `docs/OPT_SPRINT_NEXT.md` milestone status (if applicable)
- [ ] Documented any new config knobs + defaults
- [ ] Called out any format change in PR description (pre-alpha OK)

### Rollout / Toggle Plan (if behavior-affecting)
- [x] Safe defaults (feature off or conservative threshold)
- [x] Clear knobs to disable/revert behavior quickly
- [ ] Observability: counters/logs to verify effectiveness

### DIFF
--- a/TreeDB/caching/db.go
+++ b/TreeDB/caching/db.go
@@ -4325,6 +4325,11 @@ func (db *DB) ReclaimObservedValueLogSources(ctx context.Context, ids []uint32) 
 	if db == nil || len(ids) == 0 {
 		return nil
 	}
+	gcer, hasValueLogGC := db.backend.(backendValueLogGCer)
+	leafGcer, hasLeafGenerationGC := db.backend.(backendLeafGenerationGCRunner)
+	if !hasValueLogGC && !hasLeafGenerationGC {
+		return nil
+	}
 	seen := make(map[uint32]struct{}, len(ids))
 	observed := make([]uint32, 0, len(ids))
 	for _, id := range ids {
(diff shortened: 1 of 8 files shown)


snissn:
AI review sweep complete.

Resolved/fixed items:
- cached reclaim now checks backend capabilities before forcing checkpoint work.
- protected cached value-log paths are included in the GC protection set.
- the shared checkpoint scheduler latch is no longer overwritten by this path.
- cached value-log writers are rotated past observed source IDs before reclaim.
- backend rewrite post-commit leaf generation cleanup uses a non-cancelable cleanup context so an already-committed rewrite is not reported as failed because the caller context was canceled after commit.

Intentional/accepted items:
- The public cached wrapper does not enable `ObservedSourceAssumeUnreferenced` / `ObservedSourceReclaimActive` and does not force retained-path prune. That aggressive mode is unsafe for retained cached sources; the wrapper is conservative and preserves still-retained observed sources. The backend explicit option remains available and separately tested.

Validation already run locally: `go test ./TreeDB/db -run 'TestBackendValueLogRewriteReclaimsLeafGenerationDebt|TestValueLogRewriteOnline' -count=1` and `go vet ./TreeDB/db`.

PR #1456, opened 2026-05-09
experiments: validate raw JSON TreeDB fixture

snissn:
## Objective

Prevent the JSONBench raw TreeDB storage comparison from reporting invalid compression results when stored raw JSON rows do not match the source JSONL bytes.

## Context

A local raw TreeDB artifact appeared to compress full raw JSON far better than gzip/zstd. Byte-for-byte validation showed the artifact did not preserve source row contents, so the experiment harness needs to fail closed before publishing disk-footprint numbers.

## Non-goals

- Does not change TreeDB storage, value-log rewrite, or compaction behavior.
- Does not update the checked-in benchmark report numbers.
- Does not claim a new raw JSON compression result.

## Design

- Defensively copy each scanner row before inserting into the raw key/value TreeDB fixture.
- Reopen the raw TreeDB fixture after compaction/rewrite and validate every stored value byte-for-byte against the source rows before measuring final directory usage.
- Return a validation error on the first mismatch instead of producing a misleading report.

## Scope

- Includes: `experiments/colgranule/cmd/jsonbench_compare/main.go`, `experiments/colgranule/cmd/jsonbench_compare/remaining_test.go`.
- Excludes: TreeDB storage internals, value-log rewrite policy, benchmark report regeneration.

## Correctness

- Tests run: `go test ./experiments/colgranule/cmd/jsonbench_compare`.
- Corruption/edge cases covered: raw fixture validation catches source/stored row mismatches.
- Concurrency/race coverage: not applicable; experiment harness validation only.

## Performance

- Benchmarks run: none.
- Results table: not applicable.
- Regressions: raw-report generation now rereads and validates raw rows; acceptable because this is an experiment/reporting path.

## Rollout / Toggle Plan

- Default setting: validation is always on for the raw TreeDB fixture.
- How to disable quickly: revert this harness-only change.

## Follow-ups

- Regenerate the JSONBench comparison report using the validated harness.
- Investigate why value-log rewrite can increase the validated raw fixture footprint after compaction.

---

### Scope
- [x] Single, clear objective for this PR (not a bundle)
- [x] Explicit include list (files/symbols touched): `jsonbench_compare` raw fixture load/validation and tests
- [x] Explicit exclude list (files/symbols NOT touched): TreeDB storage internals and report regeneration
- [x] No unwanted feature reintroduction (e.g. async slab writer, hard zones, ValueIndex) unless explicitly stated in the PR objective

### Safety / Correctness
- [x] Clear written-vs-durable semantics for any `*Sync` path touched: no `*Sync` path touched
- [x] No new mmap usage on mutable/truncating files
- [x] All new length fields are capped before allocation (OOM-safe): no new encoded length fields
- [x] CRC/checksum behavior is fail-closed (clean error, no panic): no CRC/checksum path touched
- [x] Any concurrency change has a defined state machine and invariants: no concurrency change

### Tests
- [x] Added deterministic regression tests for the targeted failure mode
- [x] Tests avoid sleeps where possible (use latches/hooks)
- [x] Added corruption/edge-case tests (truncation, bad headers, invalid lengths) when format/parsing changes: no format/parsing change
- [x] Ran locally:
  - [ ] `go test ./... -count=1`
  - [x] Any targeted packages/benchmarks: `go test ./experiments/colgranule/cmd/jsonbench_compare`

### Benchmarks / Measurements
- [ ] Added or updated a benchmark relevant to the change
- [ ] Reported results in PR description (before/after, command, machine)
- [ ] Included “incompressible” (worst-case) results if compression is involved

### Docs
- [ ] Updated `docs/OPT_SPRINT_NEXT.md` milestone status (if applicable)
- [ ] Documented any new config knobs + defaults
- [x] Called out any format change in PR description (pre-alpha OK): no format change

### Rollout / Toggle Plan (if behavior-affecting)
- [x] Safe defaults (feature off or conservative threshold): validation fails closed
- [x] Clear knobs to disable/revert behavior quickly
- [ ] Observability: counters/logs to verify effectiveness

### DIFF
--- a/experiments/colgranule/cmd/jsonbench_compare/main.go
+++ b/experiments/colgranule/cmd/jsonbench_compare/main.go
@@ -422,6 +422,9 @@ func measureRawJSONTreeDB(ctx context.Context, files []string, rows int, dbDir s
 		out.RewriteValueBytes = rewriteStats.ValueBytesCopied
 		out.RewriteSourceBytes = rewriteStats.SourceBytesRequested
 	}
+	if err := validateRawJSONTreeDB(files, rows, dbDir); err != nil {
+		return out, err
+	}
 	after, afterFiles, err := directoryUsage(dbDir)
 	if err != nil {
 		return out, err
@@ -460,8 +463,9 @@ func insertRawJSONRows(files []string, rows int, db *treedb.DB, out *remainingTr
 				return errStopScan
 			}
 			out.Rows++
-			out.RawDocumentBytes += int64(len(raw))
-			if err := batch.Set(documentID(uint64(out.Rows)), raw); err != nil {
+			rawCopy := append([]byte(nil), raw...)
+			out.RawDocumentBytes += int64(len(rawCopy))
+			if err := batch.Set(documentID(uint64(out.Rows)), rawCopy); err != nil {
 				return fmt.Errorf("set raw JSON row %d: %w", out.Rows, err)
 			}
 			if out.Rows%batchSize == 0 {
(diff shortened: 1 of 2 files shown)


snissn:
Superseded by #1457, which includes the raw JSON validation changes from this PR plus the compaction fixes.

PR #1481, opened 2026-05-11
treedb: reduce random-read decode alloc churn (pass 1)

snissn:
## Summary
This is the alloc-first pass for random_read_parallel investigation.

### Changes
- Reuse pooled decode scratch in readViaMmapAppend even on cacheable grouped-frame misses.
- Transfer pooled ownership into grouped-frame cache entries and return pooled memory on eviction.
- Add checkpoint-settle queue-drain before read scans and print additional read-path cache/decode metrics in unified bench output.

### Why
Profiles showed high allocation pressure in compact leaf/frame decode helpers under read-miss-heavy parallel random reads. This removes a per-read heap allocation path in grouped-frame decode misses and improves observability for next iterations.

### Validation
- GOWORK=off go test ./TreeDB/internal/valuelog -count=1
- GOWORK=off go test ./cmd/unified_bench -count=1

### Benchmark note
In prior count=5 sweeps (same benchmark shape), this class of change reduced alloc objects with near-neutral throughput; this PR intentionally lands the low-risk alloc reduction first, and follow-up PRs will target decode-hit-rate and miss-path contention.

### DIFF
--- a/TreeDB/caching/checkpoint_test.go
+++ b/TreeDB/caching/checkpoint_test.go
@@ -302,19 +302,42 @@ func TestCachingDB_AutoCheckpoint_SizeTrigger_TrimsWAL(t *testing.T) {
 	db.maybeAutoCheckpoint(1<<20, autoCheckpointModeSize)
 
 	walDir := filepath.Join(dir, "wal")
-	stats := db.Stats()
-	if stats == nil {
-		t.Fatalf("Stats() returned nil")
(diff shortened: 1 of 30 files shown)


snissn:
Closing as superseded by #1489, which is now merged into main. The intended read-stack / grouped-read hitrate work has landed through the clean replacement PR, so keeping this older stacked PR open only preserves stale conflict/review noise.

PR #1519, opened 2026-05-14
[codex] prepare vector cosine query norms

snissn:
## Summary

Follow-on to #1518. Implements the structural prepared-query cosine path suggested after Priority 4:

- stores `cachedInvNorm` on immutable vector index nodes alongside `normSquared`
- prepares query inverse norm once per cosine query/insert vector
- threads the prepared query through greedy search, layer search, and neighbor selection
- makes the float32 cosine hot path only dot product plus inverse-norm multiply
- keeps checked compatibility wrappers for non-hot callers
- adds a direct microbenchmark for wrapper vs prepared unchecked query-to-node distance

This is intentionally stacked on #1518 because it depends on the scratch-reuse branch.

## Evidence

TDD/focused validation:
- Initial focused test run failed as expected before implementation: missing `cachedInvNorm`, `prepareFloat32CosineQuery`, prepared/unchecked helpers.
- `go test ./TreeDB/collections -run 'Test(VectorIndexInsertCachesStoredNorm|VectorIndexInsertCachesInt8StoredNorm|VectorIndexFloat32CosineSpecializationMatchesExactDistance|PrepareFloat32CosineQueryCachesInverseNorm|VectorIndexPruneLayerNeighborsUsesDistanceThenDocumentID|VectorIndexSearchLayerScratchReusesBuffers)$' -count=1`
- `go test -race ./TreeDB/collections -run 'Test(VectorIndexFloat32CosineSpecializationMatchesExactDistance|PrepareFloat32CosineQueryCachesInverseNorm|CollectionVectorIndexSearchReranksCanonicalRows|VectorIndexSearchLayerScratchReusesBuffers)$' -count=1`

Package validation:
- `go test ./TreeDB/collections -count=1`
- `git diff --check`

Local apples-to-apples comparison, 10k docs x 64 dims, base #1518 at `166394c4` vs this PR:

Build:
- #1518: `2.562s`, `2.581s`, `2.558s`; ~`42.39MB/op`, `183,287-183,288 allocs/op`
- This PR: `1.981s`, `2.004s`, `2.035s`; ~`42.73MB/op`, `183,281-183,290 allocs/op`

Incremental write:
- #1518: `2.596s`, `2.638s`, `2.660s`; ~`52.5-52.8MB/op`, `194,659-194,698 allocs/op`
- This PR: `2.032s`, `2.007s`, `1.993s`; ~`53.1-53.3MB/op`, `194,700-194,707 allocs/op`

Search:
[...]

Issue #1513, opened 2026-05-14
Optimize TreeDB vector ANN index build


snissn:
## Context

Follow-up from TreeDB vector search PR #1512. Build-only profiling shows vector index construction is dominated by graph construction, especially reverse-link pruning in `linkLayerLocked`.

Latest baseline profile artifact from local run:

- `/tmp/treedb_vector_build_profile_20260514_094255`
- 3k docs x 64 dims: `BenchmarkCollectionVectorIndexBuild` = `15.83s`
- CPU: `linkLayerLocked` ~94% cumulative, `vectorDistanceBetweenStoredNodes` ~93% cumulative
- Allocations: `searchLayerLocked` ~50MB, `linkLayerLocked` ~11.5MB

## Required workflow

- Update this issue checklist and work log as each priority starts, lands in a PR, or changes state.
- Use test-driven development for each optimization: add a focused failing test or benchmark guard first when practical, then implement.
- Iterate until tests and benchmarks confirm a good final state for the priority before opening/marking the PR ready.
- Keep each optimization isolated in a reviewable follow-on PR.

## Todo

- [x] Priority 1: Pre-score neighbor pruning so `linkLayerLocked` computes each neighbor distance once per prune instead of inside the sort comparator. PR #1514.
- [x] Priority 2: Cache vector norms or normalize vectors to avoid repeated cosine norm calculation. PR #1515.
- [x] Priority 3: Specialize hot float32 cosine distance paths while keeping generic/int8 paths intact. PR #1517.
- [x] Priority 4: Reuse build scratch buffers for `searchLayerLocked` visited sets/heaps/output slices. PR #1518.
- [ ] Priority 5: Reassess neighbor-selection heuristic and `efConstruction` after mechanical fixes, using recall benchmarks before changing algorithm semantics.

## Work log

- 2026-05-14: Opened issue from the build-profile analysis. First implementation target is priority 1 in a follow-on PR stacked on #1512.
- 2026-05-14: Priority 1 implemented in PR #1514. Build-only benchmark improved from `15.83s` to `2.65-2.72s` at 3k x 64 and from `54.6s` to `10.8s` at 10k x 64. Post-change profile: `/tmp/treedb_vector_build_prescore_stack_profile_20260514_095213`.
- 2026-05-14: Starting priority 2 in a new follow-on PR stacked on #1514. Requirement: TDD first, then iterate until build/query benchmarks and tests confirm the final state.
[...]

PR #1521, opened 2026-05-14
[codex] try kelindar search benchmark

snissn:
## Summary

Experimental follow-on to #1519 testing `github.com/kelindar/search`.

The public kelindar API is a full exact in-memory vector index, not a reusable public dot/cosine kernel for TreeDB's HNSW graph path. I added an optional benchmark behind `-tags kelindarbench` rather than wiring it into production TreeDB search.

## Result

This PR is draft because the candidate could not be benchmarked in this checkout without extra native runtime setup.

Evidence:
- Normal TreeDB package validation passes with the optional benchmark excluded.
- Running the kelindar benchmark with `-tags kelindarbench` panics at package init because `libllama_go.so` is missing. The panic occurs on import of `github.com/kelindar/search`, before the benchmark can run.

Observed failure:

```text
panic: Library 'libllama_go.so' not found
...
github.com/kelindar/search.init.0()
    .../github.com/kelindar/search@v0.4.1/loader.go:30
```

Validation:
- `go test ./TreeDB/collections -count=1`
- `git diff --check`
- `TREEDB_VECTOR_BENCH_DOCS=10000 TREEDB_VECTOR_BENCH_DIMS=64 go test -tags kelindarbench ./TreeDB/collections -run '^$' -bench '^BenchmarkKelindarVectorSearchExact$' -benchtime=100x -count=1 -benchmem` fails with the `libllama_go.so` init panic above.

Part of #1513.

### DIFF
--- a/TreeDB/collections/vector_search_kelindar_bench_test.go
+++ b/TreeDB/collections/vector_search_kelindar_bench_test.go
@@ -1,0 +1,30 @@
+//go:build kelindarbench
+
+package collections
+
+import (
+	"testing"
+
+	kelindarsearch "github.com/kelindar/search"
+)
+
+func BenchmarkKelindarVectorSearchExact(b *testing.B) {
+	docs := vectorBenchmarkDocs(b)
+	dims := vectorBenchmarkDims(b)
+	index := kelindarsearch.NewIndex[int]()
+	for i := 0; i < docs; i++ {
+		vector := vectorBenchmarkEmbedding(i, dims)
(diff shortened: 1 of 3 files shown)


snissn:
Closing this stale vector-library benchmark experiment.

The PR body records that the optional kelindar benchmark could not run in this checkout because importing the package required `libllama_go.so`. It remained draft and is not part of the current production column-store/native-reader mergeability work.

PR #1523, opened 2026-05-14
[codex] try gonum f32 dot

snissn:
## Summary
- replace the TreeDB float32 cosine dot loops with gonum/blas32.Dot in query-to-node and node-to-node cosine distance paths
- keep the existing prepared inverse-norm structure from #1519
- use the repo's existing Gonum v0.16.0 module version, so this does not raise the go directive

## Benchmark evidence
Baseline is #1519 / codex/treedb-vector-prepared-cosine-query on the same host.

Distance microbench, 4096 candidates x 64 dims, benchtime=200ms, count=3:
- baseline prepared_unchecked: ~111-114 us/op
- this PR prepared_unchecked: 53.111, 54.735, 53.617 us/op

10k docs x 64 dims build/write, benchtime=1x, count=3:
- baseline build: 1.981, 2.004, 2.035 s/op
- this PR build: 1.378, 1.348, 1.358 s/op
- baseline incremental write: 2.032, 2.007, 1.993 s/op
- this PR incremental write: 1.406, 1.403, 1.400 s/op

10k docs x 64 dims search, benchtime=100x, count=3:
- baseline integrated search: 1.120, 1.103, 1.117 ms/op
- this PR integrated search: 1.092, 1.128, 1.102 ms/op
- baseline graph-only search: 104, 101, 106 us/op
- this PR graph-only search: 68.699, 64.455, 66.860 us/op

Profile evidence:
- /tmp/treedb_vector_gonum_profile/build_cpu.pprof
- /tmp/treedb_vector_gonum_profile/build_mem.pprof
- build profile shows gonum.org/v1/gonum/internal/asm/f32.DotUnitary as the dot kernel after replacing both query-node and node-node dot paths

## Validation
- go test ./TreeDB/collections -run 'Test(VectorIndexFloat32CosineSpecializationMatchesExactDistance|CollectionVectorIndexSearchReranksCanonicalRows)$' -count=1
- go test ./TreeDB/collections -count=1
- git diff --check

PR #1524, opened 2026-05-14
[codex] add parallel vector graph search benchmark

snissn:
## Summary
- add BenchmarkCollectionVectorIndexGraphOnlySearchParallel using b.RunParallel
- keeps the benchmark on the graph-only path so it directly exercises concurrent query traversal over the in-memory index
- records docs, dims, and index_bytes like the serial graph-only benchmark

## Local benchmark evidence
Command:

TREEDB_VECTOR_BENCH_DOCS=10000 TREEDB_VECTOR_BENCH_DIMS=64 go test ./TreeDB/collections -run '^$' -bench '^BenchmarkCollectionVectorIndexGraphOnlySearch(Parallel)?$' -benchtime=100x -count=3 -benchmem

Results:
- serial graph-only: 67.877, 64.827, 68.633 us/op
- parallel graph-only: 16.544, 16.675, 16.906 us/op

This confirms independent graph-only searches can run concurrently through the existing RLock plus scratch-pool path.

## Validation
- go test ./TreeDB/collections -count=1
- git diff --check

Stacked on #1523.

### DIFF
--- a/TreeDB/collections/vector_search_bench_test.go
+++ b/TreeDB/collections/vector_search_bench_test.go
@@ -274,6 +274,46 @@ func BenchmarkCollectionVectorIndexGraphOnlySearch(b *testing.B) {
 	}
 }
 
+func BenchmarkCollectionVectorIndexGraphOnlySearchParallel(b *testing.B) {
+	docs := vectorBenchmarkDocs(b)
+	dims := vectorBenchmarkDims(b)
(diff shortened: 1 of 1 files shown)


snissn:
Landed on main via direct stack merge commit e00cb3cb0249e9ef7c70603620986bbdb27521c5.

PR #1526, opened 2026-05-14
[codex] cache vector graph edge distances

snissn:
## Summary
- store cached distances alongside in-memory vector graph edges
- use cached edge distances during neighbor pruning instead of recomputing node-to-node distance for every existing neighbor
- persist edge distances in vector index snapshots, with load-time fallback to recompute distances for older snapshots that only have neighbor IDs
- update memory accounting for the larger edge representation

## Local benchmark evidence
Command:

TREEDB_VECTOR_BENCH_DOCS=10000 TREEDB_VECTOR_BENCH_DIMS=64 go test ./TreeDB/collections -run '^$' -bench '^BenchmarkCollectionVectorIndex(Build|IncrementalWrite)$' -benchtime=1x -count=3 -benchmem

Results on this branch:
- build: 1.127657031s, 1.138818371s, 1.127753008s
- incremental write: 1.171219140s, 1.175518032s, 1.153284655s
- index bytes: ~8.95MB, up from ~5.96MB before cached edge distances

Previous stacked branch #1525 local reference:
- build: 1.304699088s, 1.327323916s, 1.309687977s
- incremental write: 1.371825545s, 1.353806112s, 1.344914819s

Search check:
- integrated search: 1.098942ms, 1.095098ms, 1.094685ms
- graph-only search: 63.874us, 64.616us, 67.189us
- parallel graph-only search: 15.701us, 16.177us, 17.380us

Profile artifact:
- /tmp/treedb_vector_edge_distance_profile/build_cpu.pprof
- /tmp/treedb_vector_edge_distance_profile/build_mem.pprof

Profile note:
- pruneLayerNeighborsLocked dropped to ~0.12s cumulative in the build profile after removing repeated distance scoring from that path.

## Validation
- go test ./TreeDB/collections -run 'TestVectorIndexPruneLayerNeighborsUsesDistanceThenDocumentID|TestVectorIndexSearchLayerScratchReusesBuffers|TestVectorIndexSnapshot|TestVectorIndexPersist' -count=1
- go test ./TreeDB/collections -count=1
- git diff --check

Stacked on #1525.

PR #1528, opened 2026-05-14
[codex] spec native ANN vector index

snissn:
## Summary
- add a draft spec for making the collection ANN graph a native persisted TreeDB secondary index
- cover durable keyspaces, binary record model, runtime graph loading, write semantics, recovery, rebuild/GC, concurrency, compression, observability, and phased implementation
- link the proposal from the canonical TreeDB spec README

## Validation
- go test ./TreeDB/docs -count=1
- git diff --check

Stacked on #1526 because the proposal references the current cached edge-distance graph shape.


snissn:
Landed on main via direct stack merge commit e00cb3cb0249e9ef7c70603620986bbdb27521c5.

Issue #1529, opened 2026-05-14
TreeDB: compatibility-breaking user-command WAL via typed commitlog frames


snissn:
## Summary

Implement TreeDB's replacement WAL track as a compatibility-breaking user-command WAL with applied-LSN checkpointing.

This ticket is the detailed implementation plan. PR #1530 owns the normative spec updates and should reference this ticket rather than duplicating the full execution plan in docs.

Spec paths:

- `TreeDB/docs/spec/user-command-wal.md`
- `TreeDB/docs/spec/verification.md`
- `TreeDB/docs/spec/recovery.md`
- `TreeDB/docs/spec/storage-format.md`
- `TreeDB/docs/spec/write-path-and-durability.md`

Planning PR:

- #1530
- branch: `codex/user-command-wal-spec`

## Milestone Checklist

Checked items mean the milestone is merged, issue-only administrative work is complete, or the open implementation PR is live-stack-mergeable. For implementation PRs, live-stack-mergeable means:

- the PR is open and not draft, or already merged;
- the PR has no merge conflicts with its configured base;
- every earlier PR in the active stack is also checked;
- every task, TDD evidence item, exit criterion, performance gate, evidence block, and acceptance artifact required by this tracker is complete;
- required CI/checks are green on the latest head commit;
- required local validation commands from the PR evidence block have been rerun after the latest code change, or the PR explains why a command is not applicable;
- performance gates have passed with recorded artifacts, including mode proof that command WAL was actually active when benchmark evidence claims command-WAL coverage;
- Codex and Copilot AI reviews have been requested/refreshed after each meaningful state push while building the PR;
- the latest-head Codex and Copilot reviews are passing: no unresolved actionable comments, no requested changes, and no stale "needs follow-up" comments against the latest commit;
- all prior AI review comments have either been fixed in code, answered with rationale, or become outdated/non-actionable, and actionable PR review threads are resolved;
- the PR evidence block and artifacts required by this tracker are present.

A checked implementation item is not approval to merge. Do not merge any command-WAL stack PR until the whole active stack is checked and a human gives explicit merge approval in the thread.

- [x] PR 0: spec, tracker, and deprecation cleanup — #1530
- [x] PR 1: typed commit-log frames and feature gate — #1533
[...]

PR #1530, opened 2026-05-15
[codex] Document command WAL contract and TDD tracker

snissn:
## Summary

Refs #1529.

This PR documents the approved user-command WAL direction and deprecates the previous collection root-delta WAL plan as the active implementation target.

The detailed implementation/TDD execution plan now lives in #1529. The spec in this PR owns the durable correctness contract, command support matrix, assertion categories, and verification gates.

## What Changed

- Adds `TreeDB/docs/spec/user-command-wal.md` as the canonical command WAL contract.
- Defines command WAL as a compatibility-breaking extension/replacement of the existing commit-log WAL payload format.
- Specifies raw KV as `RawKVBatch` command frames rather than legacy `commitlog.Record` compatibility payloads.
- Requires one shared WAL substrate using `wal/commit-l*.log`, with no active `collection-l*.log` or `internal/collectionwal` implementation path.
- Adds correctness requirements for root publish, `AppliedLSN` metadata atomicity, recovery restart, contiguous LSN advancement, replay idempotency/fail-closed behavior, and runtime assertion categories.
- Updates recovery, storage-format, backup/restore, value-log lifecycle, write-domain, contracts, verification, architecture, and column-store docs to point at the command WAL target.
- Points `user-command-wal.md` and `verification.md` at #1529 as the detailed implementation plan.
- Adds an existing-test migration gate so legacy WAL/recovery tests are mapped to command-WAL equivalents before deletion or rename.

## Why

The previous collection-specific root-delta WAL direction created too much complexity and risked becoming a second WAL system. The replacement direction keeps WAL semantics aligned with the existing raw WAL system while allowing all mutating surfaces, including raw KV, collections, catalog operations, and future native-wire commands, to use one typed command stream.

## Validation

- `git diff --check`
- `GOWORK=off go test ./TreeDB/docs`
- `GOWORK=off go test ./TreeDB/...`

## Notes

This is a docs/spec PR only. It does not implement the command WAL runtime.


snissn:
Planning review feedback from the column-store/WAL angle. I agree with the direction, but I would like these addressed by editing the existing specs/documents in place rather than adding a new addendum section. The goal should be that a future reader sees one coherent active plan, not old root-delta language plus later corrections.

1. Column-store RFC still has active stale `collection WAL` / root-delta language.

PR #1530 updates one column-store paragraph, but the RFC still has active wording like “must match the collection WAL transaction’s side-ref”, “write publishes, in one collection WAL transaction”, “collection WAL side ref”, “root-delta values”, and “collection WAL transaction classes”. Since column store is a major consumer of this WAL pivot, those sections should be rewritten in place to command-WAL terms or clearly marked as deprecated/historical within the existing relevant sections.

2. External-ref preparation should be explicit in the normal write path.

`user-command-wal.md` defines `ExternalRefs`, but the normal write path jumps from canonical payload to LSN/WAL append/apply. For raw KV large values, future column files, dictionaries, filters, etc., the spec should say explicitly that required external refs are prepared/written/protected before the command frame becomes recoverable, and that failure to prepare/protect them fails before visibility.

3. V1 should choose the stricter visibility model.

The docs currently allow two shapes: command execution publishes roots plus `AppliedLSN` before success, while the write-domain doc also allows mutable/queued/publishing state to become visible after the WAL frame is committed. For V1 I would make this stricter: no WAL-backed pending overlay yet; return success only after root publish plus `AppliedLSN`. Add durable pending overlays later as a separate feature after the command WAL substrate is proven.

4. `AppliedLSN` storage should be chosen before implementation PRs.

The spec correctly requires `AppliedLSN` to be selected atomically with roots, but leaves meta-page vs system-root storage open. That is okay for this planning PR, but PR1/PR2 should not proceed with both options alive. My preference is a gated meta-page field such as `AppliedCommandLSN`, because it makes recovery selection simpler and reduces split-state risk.

5. Reuse/align with native-wire deterministic command schemas.

The plan introduces `CommandEnvelope`, while native-wire already has deterministic command-entry schemas. The implementation plan should say the local command WAL payload should reuse or wrap the same canonical command payload schemas where possible. Otherwise we risk two subtly different encoders for local WAL and future Raft/native-wire commands.

6. Batch atomicity should be sharper.

V1 includes raw KV batch, collection insert batch, delete batch, and replace batch. The specs should say whether each batch is one command with one LSN and all-or-nothing semantics. If oversized batches split, they are not atomic unless there is an explicit command-group protocol.

Overall: I like the one-WAL-substrate command-WAL direction and think it is much better for column store than root-delta WAL. I would just tighten these points by editing the active docs directly so the final merged spec is internally consistent.

snissn:
Resolved in e80b2aa1.

What changed:

- Rewrote the active column-store RFC sections in place from collection-WAL/root-delta/side-ref language to command-WAL/external-ref language.
- Made external-ref preparation explicit in the normal command WAL write path: required refs are prepared, synced/protected, declared, and fail before visibility if unavailable.
- Chose the stricter V1 visibility model: no WAL-backed pending overlay; success/visibility waits for root publication plus `AppliedLSN`/`AppliedCommandLSN`.
- Chose `AppliedCommandLSN` as the V1 target storage in the gated meta-page format before implementation PRs.
- Added native-wire alignment: local command WAL payloads reuse or wrap native-wire deterministic command schemas where possible, with golden-fixture expectations.
- Sharpened batch semantics: V1 batches are one command frame, one LSN, all-or-nothing; oversized batches reject unless the caller explicitly submits smaller independent batches or a future `CommandGroup` protocol exists.

Also updated tracker #1529 so the issue body matches these decisions and does not carry the older ambiguity forward.

Local validation:

- `git diff --check`
- targeted grep for the stale phrases called out in the review comment

snissn:
Resolved the AI review feedback in the current branch head.

Changes made:
- Replaced remaining command-WAL external-ref terminology drift in backup/restore and value-log lifecycle docs.
- Removed stale system-root `AppliedLSN` storage wording from write-path/durability and pointed to the meta-page `AppliedCommandLSN` target.
- Clarified logical `AppliedLSN` vs physical `AppliedCommandLSN` wording in the verification matrix.
- Renamed the legacy raw payload section so it is explicitly pre-command-WAL.
- Fixed the deprecated collection-WAL plan typo.
- Corrected the V1 ack boundary: ordinary WAL-on visible ack requires recoverable command WAL plus normal-executor apply; root + `AppliedLSN` is reserved for flushed/synced/checkpoint/recovery cleanup boundaries.
- Added the concrete `AppliedCommandLSN` meta-page extension: body offset 60 / page offset 76, `u64`, command-WAL meta body size 68 bytes.
- Clarified PR1 only adds `RawKVBatch` payload encoding/fixtures; production raw KV conversion waits for PR3 recovery/`AppliedCommandLSN` plumbing.

Validation:
- `git diff --check`
- `GOWORK=off go test ./TreeDB/docs`

All AI review threads have been marked resolved.

snissn:
Resolved the latest Codex comment on pending-overlay tests.

Change:
- Updated the verification matrix so V1 process visibility is tested as recoverable WAL plus normal-executor install, not root publication.
- Replaced `FrameDurableButUnpublishedNotVisible` / `NoPendingOverlayBeforeRootPublish` with `FrameDurableBeforeExecutorInstallNotVisible` / `VisibleAckRecoverableBeforeCheckpoint`.
- Updated #1529 to mention visible-ack recoverability before checkpoint as part of raw command WAL evidence.

Validation:
- `git diff --check`
- `GOWORK=off go test ./TreeDB/docs`

snissn:
@codex review

snissn:
@codex review

PR #1531, opened 2026-05-15
Optimize TreeDB random read leaf cache views

snissn:
## Summary
- add a leased read-only leaf page cache view path for TreeDB point reads so cached outer-leaf pages can be searched without copying 4 KiB into scratch
- add Node view init helpers and hot-path tree load/validate Into variants to reduce node value copies
- add coverage proving cache view leases block slot mutation until release

## Profile notes
Original random_read_parallel profile showed runtime.memmove at 24.41% and runtime.duffcopy at 14.07%. After this change, the comparable all-test rerun showed Random Read (Parallel) at 6.11M ops/s, runtime.memmove at 9.52%, and no runtime.duffcopy samples in that section. Remaining memmove is from read-miss cache stores and compact leaf-log decode rather than cache-hit page copies.

After profile dir: /tmp/gomap_full_profile_after_1778805658

## Tests
- git diff --check
- go test ./TreeDB/db ./TreeDB/tree ./cmd/unified_bench -count=1

### DIFF
--- a/TreeDB/collections/api.go
+++ b/TreeDB/collections/api.go
@@ -13997,6 +13997,12 @@ func (c *Collection) bufferDirectUpdateBatchPlanLocked(plan *updateBatchPlan) (b
 			return false, ErrConcurrentMutation
 		}
 		currentBufferedPlan := plan.bufferedBase && plan.bufferedReadGeneration == domain.writeGeneration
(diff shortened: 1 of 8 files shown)


snissn:
@codex review

snissn:
@codex review

PR #1560, opened 2026-05-15
Use TreeDB profiles in vector search demo

snissn:
## Summary
- add `-profile durable|fast|wal_on_fast|bench` to `treedb_vector_search_demo`
- default the demo to TreeDB `bench` profile so benchmark runs use first-class profile storage defaults
- open through TreeDB profile-aware backend helpers, including cached leaf-log wiring for outer-leaf value-log formats
- normalize public TreeDB root layout storage reporting so `maindb/leaf_vlog`, `maindb/value_vlog`, and `maindb/index.db` are reported as first-class storage domains
- add demo-local storage knobs:
  - `-value-pointer-threshold`, default `1024`, so normal vector documents and native vector-index records stay inside outer leaves instead of spilling to `value_vlog`
  - `-leaf-generation-segment-target`, default `4194304`, so the vector demo creates sealed leaf generations that `CompactStorageFull` can pack/rewrite/GC when compaction is requested
- make the default `-require-leaf-vlog-bytes` path pass under the benchmark profile
- make the command entrypoint default to a three-case storage/search matrix: index.db outer leaves, leaf_vlog before compaction, and leaf_vlog after compaction
- make `-compact=false` the default; the matrix only compacts the explicit `leaf_vlog_after_compact` case
- replace the mistaken parallel document-read stage with parallel ANN search benchmarks via `-queries` and `-search-concurrency`

## Validation
- `GOWORK=off go test ./cmd/treedb_vector_search_demo -count=1`
- `GOWORK=off go vet ./cmd/treedb_vector_search_demo`
- `git diff --check`
- `GOWORK=off go run ./cmd/treedb_vector_search_demo -dir /tmp/treedb_vector_search_matrix_search_smoke -keep-dir -docs 1000 -dims 32 -queries 200 -search-concurrency 2,4,8 -validate-queries 4 -validate-docs 4 -top-k 5 -m 8 -ef-construction 64 -ef-search 64 -min-recall 0.5 -json >/tmp/treedb_vector_search_matrix_search_smoke.json`
- `GOWORK=off go run ./cmd/treedb_vector_search_demo -dir /tmp/treedb_vector_search_matrix_search_10k -keep-dir -docs 10000 -dims 64 -queries 10000 -search-concurrency 2,4,8,16,32,64,128 -validate-queries 16 -validate-docs 16 -top-k 10 -m 16 -ef-construction 128 -ef-search 128 -min-recall 0.95 -json >/tmp/treedb_vector_search_matrix_search_10k.json`
[...]

PR #1565, opened 2026-05-15
Use NumKong for vector cosine dot products

snissn:
Follow-up to #1560. Targets `codex/vector-demo-profile-storage` so this stays stacked on the vector search demo work.

Summary:
- replace the `gonum/blas32` cosine dot calls in the vector index graph path with NumKong `DotF32`
- preserve the cached inverse-norm path instead of switching to full cosine distance recomputation
- pin NumKong to the pseudo-version that contains the Go package

Validation:
- `GOWORK=off go test ./cmd/treedb_vector_search_demo`
- `GOWORK=off go test ./TreeDB/collections`

Benchmark note:
- Comparing `/tmp/treedb_vector_search_demo_profile_10k_deep_compact_RDOcNb.json` against `/tmp/treedb_vector_search_demo_profile_10k_deep_compact_Ry3XRQ.json`, average search latency improved from `1842.7 us` to `804.6 us` (`2.29x`) with recall `1.0`, avg candidates/rerank both `128`, and no exact fallbacks.

Review follow-up:
- Isolated NumKong behind cgo build tags with a pure-Go dot-product fallback for non-cgo builds.
- Verified the fallback with `GOWORK=off CGO_ENABLED=0 go test -run 'TestVectorIndexFloat32Cosine|TestVectorDistanceBetweenFloat32NodesCosine' -count=1 ./TreeDB/collections`.
- Added explicit mismatched-dimension coverage for `vectorDistanceBetweenFloat32NodesCosine`.
- Annotated `github.com/unum-cloud/usearch/golang` in `go.mod` as an optional `usearch_bench` benchmark dependency.


snissn:
@codex review

snissn:
Review-thread cleanup pass for current head.

The substantive comments are addressed now:

- NumKong is opt-in only with `//go:build numkong && cgo`; the default build uses the pure-Go kernel.
- `dotProductFloat32` has the same equal-length precondition and panic behavior in both pure-Go and NumKong builds.
- The helper contract documents float64 accumulation / accumulator-compatible results.
- `usearch/golang` remains in the indirect block; it is only used by the `usearch_bench` benchmark tag.

Validation:

```sh
go test -tags='numkong' ./TreeDB/collections -run 'Test(CollectionVector|Vector)' -count=1
```

Both pass locally; the tagged run emits only the linker duplicate `-lm` warning.

PR #1637, opened 2026-05-19
[codex] Add column store quickstart

snissn:
## Summary

Adds a standalone TreeDB column-store quickstart under `examples/column_store_quickstart`.

The walkthrough writes a small activity-event stream, stores `time_us`, `action`, and `actor` in physical column lanes, reopens the database, reads one row back by id, and runs two explicit physical column scans:

- `events_by_action`
- `active_window_by_actor`

The terminal output is framed with narrative text, boxed sections, and readable scan counters so the example feels like a motivated use case rather than a raw harness.

## Validation

- `GOWORK=off go run .` from `examples/column_store_quickstart`
- `GOWORK=off go test ./examples/column_store_quickstart`
- `git diff --cached --check` before commit

### DIFF
--- a/examples/column_store_quickstart/README.md
+++ b/examples/column_store_quickstart/README.md
@@ -1,22 +1,23 @@
 # TreeDB Activity Event Column Store
 
-Activity streams are a natural fit for a collection that needs both primary
-reads and scan-friendly analytics. In this walkthrough, a feed service receives
-events from web, iOS, and Android clients. Each event records when something
-happened, who acted, what action they took, which client sent it, and which feed
-item was touched.
+This command models a small feed-service activity stream. The service receives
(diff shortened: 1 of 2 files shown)


snissn:
Closing this as stale/deferred rather than retargeting it through the current production stack.

Read-only validation against current `origin/main` shows the branch is still based on the older M15C column-store stack and would be destructive if retargeted: it removes current production column-store/query/vector graph files, docs, demos, scripts, and tests while adding the standalone `examples/column_store_quickstart` material. The quickstart itself may still be useful, but it should be re-extracted as a fresh small docs/example PR on current `main` if we want it.

This keeps the merge-prep batch focused on current production code and avoids carrying a stale draft branch through the stack.

PR #2604, opened 2026-06-11
TreeDB: persist value-log ref counts on close

snissn:
## Summary

- persist the incremental value-log ref-count tracker during `DB.Close`
- prevents graceful restarts from discarding the latest ref-count metadata and forcing a full index/value-log pointer scan on the next open
- adds a regression test that writes value-log pointers, closes, verifies the persisted commit seq/counts, and reopens from the persisted tracker

## Why

During the merged geth/TreeDB Sepolia validation, a 3h run reached ~94% state / ~61% chain sync and shut down cleanly. Reopening that fresh DB then spent startup time in:

```text
TreeDB.Open -> db.initValueLogRefTracker -> scanValueLogRefCounts -> collectValueLogRefCounts
```

The captured startup CPU profile was ~99.6% in `collectValueLogRefCounts`, dominated by iterator value-log reads/LZ4 decode. The root cause is that the tracker is updated incrementally after commits but was not persisted on normal close, leaving stale on-disk metadata for the next open.

## Tests

```sh
go test ./TreeDB/db -run 'TestValueLogGC_IncrementalTrackerPersistsOnClose|TestValueLogGC_IncrementalParityWithFullScan|TestValueLogGC_IncrementalMode_RebuildOnCorruption' -count=1 -v
go test ./TreeDB/db ./TreeDB/tree ./TreeDB/zipper -count=1
go test ./TreeDB/... -count=1
```

## Caveats

- This optimizes graceful close/reopen. Ungraceful exits can still require rebuilding missing/stale metadata.
- TreeDB remains experimental for geth; Pebble/default behavior is unchanged.

### DIFF
--- a/TreeDB/db/db.go
+++ b/TreeDB/db/db.go
@@ -1805,6 +1805,11 @@ func (db *DB) Close() error {
 	db.closing.Store(true)
 	db.stopCommitCombiner()
 	db.pruner.Stop()
+	if db.valueLogRefTracker != nil {
+		if err := db.persistValueLogRefTracker(); err != nil {
+			errs = append(errs, err)
+		}
+	}
 	if db.ghostManager != nil {
 		db.ghostManager.stop()
 	}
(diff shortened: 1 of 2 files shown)


snissn:
@codex review
@copilot review

Issue #3040, opened 2026-06-25
TreeDB Raft R3a-2: deterministic apply harness and rejection boundary


snissn:
## Goal

Implement the R3a deterministic apply harness skeleton and fail-closed rejection boundary.

Done means deterministic entry bytes can be decoded, validated, classified, rejected before local command-WAL append, and reported with deterministic `ApplyResultV1` statuses, while accepted command lowering remains owned by later slices.

## Parent

Parent tracker: #1654

## Scope

- [ ] Add an internal apply API for committed deterministic entry bytes.
- [ ] Accept explicit apply metadata such as `ApplyEntryID` and selected local durability boundary.
- [ ] Decode and validate entries from deterministic bytes, not native-wire request structs.
- [ ] Reject unsupported entries before local WAL append, `AppliedLSN`, visible mutation, result/idempotency records, or apply-progress advancement.
- [ ] Add fake/test `ApplyProgressStore` semantics if real persistence is deferred.
- [ ] Enforce bounded fake/test progress and result stores.
- [ ] Expose stable result/error names for rejected unsupported, malformed, conflict, unsafe durability, and read-only cases.

## Non-Goals And Boundary

- [ ] Do not implement real networked Raft.
- [ ] Do not report `raft_committed` success.
- [ ] Do not widen the accepted command set beyond the R3a allowlist.
- [ ] Do not implement Mongo gateway routing here.

## Required Tests

- [ ] Rejected unsupported entry emits no command-WAL frame.
- [ ] Rejected unsupported entry does not advance `AppliedLSN` or apply progress.
- [ ] Rejected unsupported entry does not write idempotency/result records.
- [ ] Rejected unsupported entry does not mutate catalog or collection state.
- [ ] Malformed bytes return deterministic errors without panic.
- [ ] Harness rejects native-wire request objects or handler replay as input.
- [ ] Bound overflow fails before local WAL append or visible mutation.
- [ ] Read-only DB apply fails without mutation.

## Benchmarks And Evidence

- [ ] Benchmark rejected-entry decode/validation if the new path affects hot code.
- [ ] Report `ns/op`, `B/op`, `allocs/op`, payload size, rejection reason counts, and whether local command-WAL append count stayed zero.
- [ ] Before/after evidence is required for any optimization claim.

## Branch And PR Policy

- [ ] Topic branch from current `main`.
- [ ] PR body includes start-phase test plan, close-phase test evidence, benchmark evidence or rationale, and #1654 linkage.
[...]

Issue #3038, opened 2026-06-25
TreeDB Raft R3a-1: CommandEntryV1 digest result and allowlist contracts


snissn:
## Goal

Define the R3a deterministic command-entry contract before any command is accepted by the local Raft apply harness.

Done means `CommandEntryV1`, `CommandDigestV1`, `ApplyResultV1`, target identity, replication-scope rule, idempotency mode, and the R3a allowlist/rejection matrix are implemented or documented as executable internal contracts with drift tests and fixtures.

## Parent

Parent tracker: #1654

## Current Evidence

- Native-wire already has deterministic entry encoding in `TreeDB/internal/nativewire/deterministic.go`.
- Current replicated native-wire schemas include create collection, create index, drop index, insert batch, replace batch, delete batch, and `update_bson_set`.
- `TreeDB/docs/spec/command-wal-nativewire-alignment.json` shows create/drop index are deterministic at native-wire level but still `WAL-rejected` locally.
- `ack_policy` and transport metadata are not deterministic command identity.

## Scope

- [ ] Define or confirm `CommandEntryV1` bytes for R3a-supported commands.
- [ ] Define mandatory `CommandDigestV1` coverage.
- [ ] Define deterministic `ApplyResultV1` statuses and result payload/digest rules.
- [ ] Define target identity: database/catalog scope, collection identity/generation, and schema/catalog guard.
- [ ] Define literal `single-group-v1` or schema-versioned single-group rule for R3a.
- [ ] Define idempotency mode and ban production use of `NoIdempotencyV1`.
- [ ] Define resource bounds for entry size, item count, result/idempotency records, and apply-progress records.
- [ ] Define command classification rows: replay-required, duplicate-fail-closed-allowed, rejected.
- [ ] Add or update fixtures proving digest stability and metadata invariance.

## Commands In Scope

Contract-only classification for:

- create collection
- insert batch
- replace/update-by-id
- delete-by-id
- `update_bson_set`

## Commands Explicitly Out Of Scope Or Rejected

- create/drop index until local command-WAL payloads and recovery tests land
- drop collection
- flush, checkpoint, and maintenance barriers
- ordinary reads and cursor state
- arbitrary Mongo query/update/delete shapes

## Required Tests

- [ ] Supported commands appear in the allowlist with expected command-WAL status.
- [ ] `WAL-rejected` or local-only commands fail the R3a allowlist.
[...]

Issue #1654, opened 2026-05-20
TreeDB Raft R3a: deterministic command-entry apply harness over user-command WAL


snissn:
## Goal

Land the TreeDB Raft R3a state-machine boundary for deterministic application command entries over the landed user-command WAL substrate.

This issue is explicitly **not** the full networked Raft implementation. The target is the in-process apply harness that future Raft consensus will call after a log entry is committed:

```text
deterministic replicated command-entry bytes
-> decode without reconstructing native-wire requests
-> validate command kind/version/catalog/idempotency metadata
-> lower supported entries to local user-command WAL payloads
-> append/apply through the normal TreeDB command-WAL executor path
-> publish local effects and AppliedLSN according to local durability rules
-> record or simulate apply-progress/idempotency/result ordering without outrunning local recoverability
-> prove logical convergence and deterministic outcomes across fresh DBs
```

The local command WAL remains node-local crash recovery state. The Raft log remains replicated consensus ordering. R3a builds the boundary between those two systems before choosing or wiring a Raft library.

## Execution Ordering And Blocking

Milestones must execute in order. Later behavior may not be exposed early because a schema, benchmark label, or placeholder API exists.

- R3a-0 owns cleanup and documentation alignment. Old physical/root-delta collection WAL work is historical only. New work must reference the command-envelope / `AppliedLSN` model.
- R3a-1 owns the deterministic command-entry contract, target identity contract, result contract, and allowlist. R3a may not accept any command that is not deterministic and `WAL-supported` in the command-WAL matrix.
- R3a-2 owns the apply harness API and rejection boundary. Unsupported entries must fail before local WAL append, `AppliedLSN` advancement, apply-progress advancement, result/idempotency advancement, or visible mutation.
- R3a-3 owns lowering supported command entries into local command-WAL payloads and applying through the normal executor. It may not add Raft-specific mutation bypasses.
- R3a-4 owns logical digest and convergence proof across fresh DBs. It may not claim replicated-state determinism from byte-identical local files, root IDs, page IDs, local WAL frame bytes, or local value-log RIDs.
- R3a-5 owns crash/recovery ordering, duplicate/idempotency behavior, result replay behavior, and fake-or-real apply-progress metadata boundaries. Durable applied-index/idempotency/result metadata must not advance past local recoverability.
- R3a-6 owns benchmark/docs/closeout. It must distinguish deterministic-entry decode/lower/apply costs from networked Raft, checkpoint, snapshot, and public API costs.

[...]

Issue #3037, opened 2026-06-25
TreeDB Raft R3a-0: child ownership map and docs alignment


snissn:
## Goal

Turn #1654 from a hardened umbrella contract into an executable child-issue work plan before code-bearing R3a implementation begins.

Done means #1654 has linked children, each child owns a bounded checklist, and TreeDB docs/tests point at the current command-envelope / `AppliedLSN` model rather than legacy physical collection-WAL language.

## Parent

Parent tracker: #1654

## Why This Exists

#1654 explicitly says it is the umbrella acceptance contract, not the executable implementation checklist. The first step is process and documentation alignment so implementation PRs close against child trackers and do not keep expanding the umbrella with PR-local detail.

## Current Evidence

- #1654 is open and already defines R3a ordering, contracts, required child issues, tests, and completion criteria.
- `TreeDB/docs/spec/native-query-raft-roadmap.md` states that user-command WAL is local crash-recovery state and not a Raft log.
- `TreeDB/docs/spec/user-command-wal.md` and `TreeDB/docs/spec/write-path-and-durability.md` define local command-WAL and `AppliedLSN` rules.
- `TreeDB/docs/spec/command-wal-nativewire-alignment.json` records current native-wire to command-WAL status.

## Scope

- [ ] Add or maintain a linked child issue list in #1654.
- [ ] Confirm each R3a child owns a distinct implementation slice.
- [ ] Confirm old physical/root-delta collection-WAL text is superseded or clearly historical in docs touched by this work.
- [ ] Keep #1654 as acceptance rollup and cross-cutting contract authority.
- [ ] Keep child issues as executable PR checklists.

## Non-Goals And Boundary

- [ ] Do not implement the apply harness here.
- [ ] Do not choose or integrate a Raft library here.
- [ ] Do not widen the replicated command set here.
- [ ] Do not claim networked Raft or `raft_committed` support from this issue.

## Execution Ordering And Blocking

This issue must close before the first code-bearing R3a implementation PR claims ownership of #1654 work. Docs-only cleanup may land here, but runtime code should land in the relevant child issue.

## Branch And PR Policy

- [ ] Work happens on a topic branch from current `main`.
- [ ] Direct pushes to `main` are prohibited unless explicitly authorized.
- [ ] PRs include linked tracker issue and milestone, start-phase test plan, close-phase test evidence, and benchmark rationale when benchmarks are not relevant.
- [ ] Latest-head CI is required for mergeability; stale green checks do not count.
[...]

PR #3056, opened 2026-06-25
TreeDB: add R3a command-WAL apply seam

### LINKED ISSUE #3039: TreeDB Raft R3a-1/R3a-2: command-WAL apply API extraction spike
## Goal

Extract or prove the minimal command-WAL append/apply/finalize APIs needed by the R3a apply harness without replaying native-wire requests or bypassing normal TreeDB executors.

Done means the harness can call an internal boundary that accepts deterministic entry bytes plus apply metadata, lowers accepted entries into local command-WAL payloads, and observes local recoverability ordering without direct mutation shortcuts.

## Parent

Parent tracker: #1654

## Current Evidence

- Local command-WAL append/publish/replay paths live under `TreeDB/db` and `TreeDB/collections`.
- Collection replay handlers are registered through `collections.RegisterCommandWALReplayHandlers()` and apply through normal collection executors.
- Native-wire requests currently execute through server request handlers, which are not an acceptable substitute for committed-entry apply.

## Scope

- [ ] Inventory current command-WAL APIs and identify what can be reused without widening public surface.
- [ ] Add the smallest internal API or adapter needed by R3a.
- [ ] Make local command-WAL append visible and testable in the apply path.
- [ ] Preserve normal collection/catalog executor use below the command-WAL/replay layer.
- [ ] Expose explicit recoverability status names: locally applied, locally WAL-recoverable, locally root-recoverable.
- [ ] Fail read-only or unsafe durability modes explicitly.

## Non-Goals And Boundary

- [ ] Do not accept new replicated commands here.
- [ ] Do not implement networked Raft here.
- [ ] Do not add public API unless an internal seam cannot satisfy the harness.
- [ ] Do not bypass local command WAL for mutations.

## Required Tests

- [ ] Internal seam can append and publish a no-op or test frame only when scoped by this issue.
- [ ] Unsupported or malformed input fails before command-WAL append.
- [ ] Read-only DB apply returns an explicit unsupported/read-only error.
- [ ] Existing direct collection/native/Mongo behavior remains unchanged.
- [ ] Command-WAL poisoning or recovery-required states propagate rather than being hidden.

## Benchmarks And Evidence

- [ ] If runtime code changes a hot path, include before/after focused benchmarks and allocation data.
- [ ] If this remains an API extraction/spike, record why no performance benchmark is meaningful and include tests proving no behavior widening.

## Branch And PR Policy

- [ ] Topic branch from current `main`.
- [ ] PR body states exact API ownership, callers, and why the boundary prevents request replay or executor bypass.
- [ ] Latest-head CI is required; stale green checks do not count.
- [ ] Request AI reviews only after tests and evidence are current.

## Completion Criteria

- [ ] R3a follow-on issues can call the seam without inventing a parallel mutation stack.
- [ ] The seam keeps local command WAL as local recoverability and does not conflate it with Raft consensus.

snissn:
## Summary
- add internal `TreeDB/internal/commandwalapply` seam for already-lowered command-WAL frames with explicit local recoverability statuses: `locally_applied`, `locally_wal_recoverable`, and `locally_root_recoverable`
- accept only a scoped empty RawKV no-op test frame in this PR, so the seam can prove append/finalize behavior without widening replicated command support
- reject unsupported/malformed input before command-WAL append and return explicit read-only errors from low-level command-WAL intent/publish helpers

## Scope / Links
Fixes #3039
Part of #1654

This PR owns only the local command-WAL apply seam. It does not define #3038's deterministic entry/result/allowlist contract, does not build #3040's apply harness, and does not accept create/insert/update/delete replicated entries.

## Tests
- `GOWORK=off go test ./TreeDB/internal/commandwalapply`
- `GOWORK=off go test ./TreeDB/db -run 'CommandWAL|ReadOnly'`
- `GOWORK=off go test ./TreeDB/collections -run 'CommandWAL|ColumnStore.*CommandWAL'`
- `GOWORK=off go test ./TreeDB/nativewire -run 'CommandWAL|Mutation|Metadata'`
- `GOWORK=off go test ./TreeDB/mongo_gateway/... -run 'CommandWAL|Mutation|Insert|Update|Delete'`
- `git diff --check`

## Benchmarks
No benchmark run. This is an internal API extraction/spike: no production write path calls the new package, and the only accepted frame is an empty scoped test no-op. Focused tests prove no accepted mutation widening and no frame append on malformed/unsupported/read-only inputs.

## Risks / Non-goals
- Future accepted mutation classes still need executor-specific integration so create/insert/update/delete entries go through normal TreeDB catalog/collection executors before finalize.
- The explicit metadata slot is intentionally empty here; #3038/#3040 must define and consume the actual deterministic entry/result/apply metadata.
- This does not implement networked Raft or replay native-wire request handlers.

### DIFF
--- a/TreeDB/db/command_wal_raw.go
+++ b/TreeDB/db/command_wal_raw.go
@@ -252,6 +252,9 @@ func (db *DB) NewCommandWALIntent(kind commitlog.CommandKind, scope commitlog.Co
 	if db == nil || !db.commandWAL {
 		return nil, nil
 	}
+	if db.readOnly {
+		return nil, ErrReadOnly
+	}
 	if db.durability == DurabilityWALOffRelaxed {
 		return nil, fmt.Errorf("%w: WAL-off durability is incompatible with command WAL", ErrCommandWALUnsupported)
 	}
@@ -563,6 +566,9 @@ func (db *DB) appendPublicCommandWALIntent(intent *CommandWALIntent, sync bool) 
 // publishing roots. It is used by cached public command-WAL writers that must
 // make a typed frame replay-visible before inserting the mutation into memory.
 func (db *DB) AppendCommandWALIntent(intent *CommandWALIntent, sync bool) (uint64, error) {
+	if db != nil && db.readOnly {
+		return 0, ErrReadOnly
+	}
 	return db.appendPublicCommandWALIntent(intent, sync)
 }
 
@@ -572,6 +578,9 @@ func (db *DB) AppendCommandWALIntent(intent *CommandWALIntent, sync bool) (uint6
 func (db *DB) AppendCommandWALPayload(kind commitlog.CommandKind, scope commitlog.CommandScope, payloadFormat commitlog.PayloadFormat, payload []byte, sync bool) (uint64, error) {
 	if db == nil || !db.commandWAL {
 		return 0, nil
+	}
+	if db.readOnly {
+		return 0, ErrReadOnly
 	}
 	if db.durability == DurabilityWALOffRelaxed {
 		return 0, fmt.Errorf("%w: WAL-off durability is incompatible with command WAL", ErrCommandWALUnsupported)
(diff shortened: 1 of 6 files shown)


snissn:
Superseded by #3061. The original branch carried the already-merged #3037 docs through its no-force-push refresh path, so #3061 recreates the same #3039 implementation from current main with a clean diff.

Issue #3630, opened 2026-07-09
Land TreeDB GC/vacuum/maintenance efficiency punchlist


snissn:
## Goal

Land the GC / vacuum / maintenance efficiency punchlist as an ordered TreeDB issue stack. Done means every `PL-xx` child below is either implemented with tests and before/after evidence, or explicitly superseded by a linked issue that owns the same measured gate.

Source analysis was refreshed against current `main` at commit `2182e84bd668f6ea610726717d90e09a86a17a32`. No on-disk format migration work is expected for this stack; TreeDB is pre-alpha, but these issues should preserve persistent value-log semantics and pointer durability.

## Child Issues

| Order | Punchlist | Issue | Primary gate |
| ---: | --- | --- | --- |
| 1 | PL-13 | https://github.com/snissn/gomap/issues/3631 | `valueLogGC` returns cleanly when the refreshed set is still nil/empty; no panic. |
| 2 | PL-05 | https://github.com/snissn/gomap/issues/3632 | Idle background vacuum ticks with unchanged `CommitSeq` perform zero page/freelist/collection walks. |
| 3 | PL-12 | https://github.com/snissn/gomap/issues/3633 | Background vacuum eventually runs under sustained backlog and can trigger on freelist/collection-root debt. |
| 4 | PL-10 | https://github.com/snissn/gomap/issues/3634 | Fenced reclaim audits use the tracker-aware referenced set or one validated shared scan, not an unconditional full scan per audit. |
| 5 | PL-08 | https://github.com/snissn/gomap/issues/3635 | Full-mode CompactStorage skips nearly-live value-log segments by policy while exhaustive mode remains byte-minimizing. |
| 6 | PL-03 | https://github.com/snissn/gomap/issues/3636 | CompactStorage audit reachability walks collapse to one shared pass per valid audit. |
| 7 | PL-02 | https://github.com/snissn/gomap/issues/3637 | Leaf-generation GC scans outside `writeMu`; foreground writes are not blocked for O(P). |
| 8 | PL-01 | https://github.com/snissn/gomap/issues/3638 | Leaf-generation pack copy happens outside `writeMu`; writers pause only for validate/publish. |
| 9 | PL-04 | https://github.com/snissn/gomap/issues/3639 | Value-log GC full reachability scan happens outside `publishPrepareMu`; deletion safety is fenced. |
[...]

PR #3664, opened 2026-07-10
treedb: move value-log GC scan outside publish lock

### LINKED ISSUE #3639: [PL-04] Move value-log GC reachability scan outside publishPrepareMu
Parent: https://github.com/snissn/gomap/issues/3630

## Goal

Fix `PL-04`: stop holding `publishPrepareMu` across the full value-log GC reachability scan while preserving deletion safety for prepared-but-not-yet-published value-log segments.

Done means `valueLogGC` scans referenced segments before the exclusive publish-prepare lock, then takes the lock only for cheap revalidation, candidate marking, value-log set publish, and deletion accounting.

## Current Evidence

`TreeDB/db/vlog_gc.go` takes `db.publishPrepareMu.Lock()` for non-dry-run GC before calling `db.referencedValueLogSegments(ctx)`. If the incremental tracker is disabled or invalid, that call scans all maintenance roots. Commit finalization and value-log append/publish preparation use the same lock family, so commits can stall for the full O(N) scan.

The comment is right that GC must serialize with prepared-publish windows; the problem is the lock covers scan work, not just the deletion-safety window.

## Required Fix

- Split `valueLogGC` into pre-lock scan and locked validate/delete phases.
- Pre-lock phase:
  - compute referenced set using tracker or full scan;
  - record scanned commitSeq and value-log set snapshot/fence inputs;
  - record protected paths and current writable/recent segment keep set.
- Locked phase under `publishPrepareMu`:
  - re-read current commitSeq/value-log set;
  - if unchanged, use the precomputed referenced set;
  - if changed, either apply a safe tracker delta, rescan under a bounded retry strategy outside the lock, or conservatively keep all segments newer than the scanned fence;
  - ensure segments appended/prepared after the scan are protected by commitSeq and recent/current-writable fences;
  - mark candidates zombie and publish value-log set as today.
- Add a written safety invariant in code comments and PR body: never zombie a segment that could be referenced by a root published after the scan.
- If the current `valueLogKeepRecentSegmentsPerLane` window is not enough to prove safety, replace or augment it with a commitSeq/fileID fence.

## TDD Plan

Required tests:

- [ ] Publish latency test: pause `referencedValueLogSegments`/full scan with a hook and assert a concurrent commit finalization is not blocked by the scan before the locked phase.
- [ ] Race safety test: create a segment after the scan and publish a root referencing it before GC locks; GC must not mark/delete that segment.
- [ ] Tracker-stale/full-scan fallback test: invalid tracker path still obeys the same safety fence.
- [ ] Existing value-log GC premature-deletion, protected-path, maintenance-root, and pinned-set tests remain green.
- [ ] Race coverage for the new split path.

Validation command:

```sh
GOWORK=off go test ./TreeDB/db ./TreeDB -run 'TestValueLogGC|TestMaintenanceRoots|TestCommandWAL|TestPublish' -count=1
GOWORK=off go test -race ./TreeDB/db -run 'TestValueLogGC' -count=1
```

## Performance Gates

| Gate | Current | Target | Required evidence | Failure action |
| --- | ---: | ---: | --- | --- |
| publish blocking | O(N) scan under `publishPrepareMu` | concurrent publish proceeds while scan is paused | deterministic latch test | block merge |
| lock hold | scan + classify + mark + publish | revalidate/mark/publish only; target >=10x lock-hold reduction on full-scan fixture | lock-duration counter or mutex profile | profile/tune before merge |
| deletion safety | lock serializes everything | commitSeq/fileID/recent/current fences prove no post-scan referenced segment is deleted | race safety tests and written invariant | block merge |
| GC effectiveness | conservative fences may keep too much | eligible old unreferenced segments still delete | before/after GC fixture | tune fence or link blocker |

Suggested benchmark:

```sh
GOWORK=off go test ./TreeDB/db -run '^$' -bench 'BenchmarkValueLogGC|BenchmarkValueLogRewriteOnline_ValuePointers' -benchmem -count=5
```

If no benchmark exists, add a focused GC benchmark that reports scan wall time, locked-phase wall time, publish p95/p99 during GC, segments eligible/deleted/kept-by-fence, `B/op`, and `allocs/op`.

## Non-Goals

- Do not weaken value-log persistence or delete-by-recent-age.
- Do not solve CompactStorage audit scan duplication here; `PL-03` owns that.
- Do not change value-log file format.

## PR Requirements

Use a topic branch. PR body must include the deletion-safety invariant, failure-mode analysis, exact tests, race evidence, and before/after publish latency/lock data. Latest-head CI and mature AI reviews are required before mergeability is claimed.

---

## Review corrections (Fable audit, 2026-07-09) — Verdict: MINOR

Citations verified accurate (exclusive `publishPrepareMu.Lock()` at `vlog_gc.go:133-134`, scan at `:141`; shared-lock readers `finalize_commit_prepare.go:76`, `value_log_appender.go:102-103`, `db.go:2554`; tracker-invalid full-scan path via outer-leaf nil tracker `vlog_gc_incremental.go:218-228`). The pre-lock-scan / locked-revalidate structure is the right fix. The following corrections are **required** before implementation:

**1. Keep-set computation is on the wrong side of the split.** "Required Fix → Pre-lock phase" says to record "protected paths and current writable/recent segment keep set" before the lock. That is wrong: `keptIDs` (`vlog_gc.go:171-186`) and especially `pendingValueLogAppendFileIDs()` (`:187-190`) protect prepared-but-unpublished segments and **must be computed inside the locked phase**, against the same set snapshot used to enumerate candidates.

**2. The "keep all segments newer than the scanned fence" fallback option is unsound on its own.** Batch entries can carry pre-built pointers (`batchpkg.Entry.IsPtr`/`ValuePtr`, see `vlog_gc_incremental.go:586`), so a root published after the scan can add a *first* reference to a segment *older* than the fence — a newer-than-fence keep set never protects it. **Strike this option** (or gate it behind a proof that no commit can first-reference a pre-fence segment). Prefer the tracker-delta option (`valueLogRefDelta.positives` already records per-commit added refs) or a bounded abort/retry.

**3. Add a `replace()`-clobber guard.** `referencedValueLogSegments` unconditionally does `tracker.replace(counts, scannedSeq, true)` after a fallback scan (`vlog_gc_incremental.go:476-477`). Moved outside the lock, a concurrent commit can advance the tracker past `scannedSeq`; `replace()` then rolls it back, silently degrading the tracker to perpetual rescans (fail-safe but wasteful). The pre-lock scan path must skip/guard `replace()` when `tracker.commitSeq > scannedSeq`.

**4. commitSeq-unchanged ≠ no-op.** Prepared-unpublished segments exist precisely in that state and are covered by pins + active/recent keeps computed under the lock, not by the referenced set. State this so the implementer doesn't treat commitSeq equality as a full safety proof.

**Added test requirement:** the race test must include an **old, unreferenced segment gaining its first reference post-scan** (the ticket's current race test only covers a *new* segment created post-scan).

snissn:
## Summary

Fixes #3639  
Parent: #3630

- move value-log GC reachability resolution before the exclusive `publishPrepareMu` phase
- fence the scanned referenced set against the root commit sequence under the lock, retry once outside the lock, and conservatively abort after a second mismatch
- compute current/recent/protected/pending-append keep sets only inside the locked phase against the current value-log topology
- prevent a fallback scan from overwriting tracker counts advanced concurrently during that scan, while preserving existing stale-ahead tracker repair
- add deterministic old-segment, new-segment, paused-scan publish, bounded-retry, tracker, race, and paired performance coverage

Base: `2c1cab846aa803aba7d7ef2c6a2026d3420da56e`  
Head: `ad9469c68e1004306be479a741fa655ec53b07c9`

## Deletion-safety invariant

GC never zombifies a segment from a referenced set older than the published root visible while `publishPrepareMu` is exclusively held.

The pre-lock scan returns both the referenced segment set and its snapshot `CommitSeq`. The locked phase may classify/delete only when the current sequence exactly matches that scanned sequence. One mismatch unlocks and retries the full operation outside the lock; a second mismatch returns without marking or deleting segments.

Sequence equality fences published roots only. Prepared-but-unpublished output can exist without changing `CommitSeq`, so the current/recent segment set and `pendingValueLogAppendFileIDs()` are intentionally read under the same exclusive lock used for candidate classification. This retains the existing prepared-output safety contract.

The implementation does not use a file-ID/newer-segment fence. A post-scan root can first-reference an old segment via a pre-built `ValuePtr`; the exact sequence fence and bounded rescan cover that case.

## Tracker contract

Fallback scans used by GC capture a private tracker revision before scanning. Replacement is skipped only when the tracker changed during the scan and now contains counts for a sequence newer than the scanned sequence. This prevents concurrent exact counts from being rolled back.

Ordinary fallback consumers retain unconditional exact-scan repair. An unchanged stale-ahead tracker can therefore still be replaced by the current scanned sequence, preserving the CompactStorage recovery contract tested by `TestCompactStorageFencedReclaimFallsBackWhenTrackerStale`.

## Failure modes

| Interleaving | Result |
| --- | --- |
| no root change during scan | short locked classification/delete phase proceeds |
| one post-scan root publication | unlock, rescan outside the lock, then revalidate |
| publications after both scan attempts | safe zero-delete abort |
| old unreferenced segment gains first reference | sequence mismatch forces rescan; segment remains readable |
| new segment is created and first referenced | sequence mismatch forces rescan; new segment remains readable |
| prepared output exists without root publication | locked current/recent/pending keep sets protect it |
| tracker advances while fallback scan runs | revision guard preserves newer exact counts |
| tracker is stale-ahead but unchanged | fallback scan repairs it to the current sequence |
| context cancellation or scan error | existing error propagation; no candidate publication |

## Regression and TDD evidence

The exact base reproduces the original lock-scope failure: while a fallback scan is deterministically paused for 25 ms, a concurrent publish waits about 25.5 ms. On the head, the same publish finishes around 1.2 ms before GC enters its exclusive phase.

The initial worker milestone committed the first behavior tests with the implementation, so no separate red execution was preserved for those tests. The subsequent full-package gate did produce and preserve a real red/green cycle: `TestCompactStorageFencedReclaimFallsBackWhenTrackerStale` failed because the first tracker guard blocked stale-ahead repair. The guard was redesigned around a private revision token; the existing regression plus new concurrent-advance and stale-repair tests then passed repeatedly and under race.

Added deterministic coverage:

- `TestValueLogGC_FullScanDoesNotBlockPublishBeforeLockedPhase`
- `TestValueLogGC_PostScanFirstReferenceToOldSegmentIsSafe`
- `TestValueLogGC_PostScanNewSegmentIsSafe`
- `TestValueLogGC_RepeatedPostScanPublicationAbortsWithoutDelete`
- `TestValueLogRefTracker_GCReplaceDoesNotRegressConcurrentAdvance`
- `TestValueLogRefTracker_GCReplaceRepairsUnchangedStaleAheadState`
- `BenchmarkValueLogGCPublishDuringPausedFullScan`

## Validation

All commands ran on the final integrated head:

```sh
GOMEMLIMIT=4GiB GOMAXPROCS=2 GOWORK=off go test -p 1 ./TreeDB/db ./TreeDB \
  -run 'TestValueLogGC|TestMaintenanceRoots|TestCommandWAL|TestPublish|TestCompactStorageFencedReclaimFallsBackWhenTrackerStale' \
  -count=1 -timeout=10m
# PASS: db 5.227s; TreeDB 0.378s

GOMEMLIMIT=4GiB GOMAXPROCS=2 GOWORK=off go test -race -p 1 ./TreeDB/db \
  -run 'TestValueLogGC|TestValueLogRefTracker_GCReplace' -count=1 -timeout=10m
# PASS: db 1.842s

GOWORK=off go test ./TreeDB/db \
  -run 'TestCompactStorageFencedReclaimFallsBackWhenTrackerStale|TestValueLogRefTracker_GCReplace|TestValueLogGC_(FullScanDoesNotBlockPublishBeforeLockedPhase|PostScanFirstReferenceToOldSegmentIsSafe|PostScanNewSegmentIsSafe|RepeatedPostScanPublicationAbortsWithoutDelete)' \
  -count=20
# PASS: db 2.794s

GOWORK=off go test -race ./TreeDB/db \
  -run 'TestValueLogRefTracker_GCReplace|TestValueLogGC_(FullScanDoesNotBlockPublishBeforeLockedPhase|PostScanFirstReferenceToOldSegmentIsSafe|PostScanNewSegmentIsSafe|RepeatedPostScanPublicationAbortsWithoutDelete)' \
  -count=10
# PASS: db 3.293s

GOMEMLIMIT=4GiB GOMAXPROCS=2 GOWORK=off go test -p 1 ./TreeDB/db -count=1 -timeout=10m
# PASS: db 245.298s

GOMEMLIMIT=4GiB GOMAXPROCS=2 GOWORK=off go vet ./TreeDB/db
git diff --check
# PASS
```

## Performance gate

Host: Linux amd64, Intel i5-11400F, Go 1.26.0. The identical benchmark file was applied to a detached exact-base worktree and committed on the head. Each side ran 20 externally alternating samples with one benchmark operation per process:

```sh
GOWORK=off go test ./TreeDB/db -run '^$' \
  -bench '^BenchmarkValueLogGCPublishDuringPausedFullScan$' \
  -benchtime=1x -count=1 -benchmem
```

The 25 ms deterministic scan pause isolates the active window. Setup and GC completion are outside the benchmark timer; the timed region is the concurrent `DB.Set`. Every base and head sample reports exactly one eligible segment deleted.

| Metric, n=20 | Base `2c1cab846a` | Head `ad9469c68` | Result |
| --- | ---: | ---: | ---: |
| publish median | 25.527 ms | 1.190 ms | 21.45x lower |
| publish p95 | 25.590 ms | 1.221 ms | 20.96x lower |
| publish p99 | 25.626 ms | 1.229 ms | 20.85x lower |
| benchmark median | 25.545 ms/op | 1.203 ms/op | 21.23x lower |
| median B/op | 67,184 | 34,960 | -47.97% |
| median allocs/op | 319 | 75 | -76.49% |
| segments deleted/op | 1.0 | 1.0 | unchanged |

This passes the required >=10x lock-window reduction at median, p95, and p99 without weakening GC effectiveness.

## Compatibility and scope

- no on-disk format or persistent metadata encoding change
- no public API change
- persistent value-log pointers and deletion eligibility semantics remain unchanged
- observed-only GC keeps its existing locked behavior and skips the reachability scan as before
- CompactStorage shared-audit work remains owned by PL-03; this PR changes only value-log GC scan/validation and tracker replacement fencing

Rollback is a normal code revert; no migration or persisted toggle is involved.

### DIFF
--- a/TreeDB/db/vlog_gc.go
+++ b/TreeDB/db/vlog_gc.go
@@ -171,7 +171,7 @@ func (db *DB) valueLogGC(ctx context.Context, opts ValueLogGCOptions, lockMainte
 		const maxStaleScanRetries = 1
 		for attempt := 0; ; attempt++ {
 			var err error
-			referenced, _, scannedSeq, err = db.referencedValueLogSegmentsWithSourceAtSeq(ctx)
+			referenced, _, scannedSeq, err = db.referencedValueLogSegmentsForGCAtSeq(ctx)
 			if err != nil {
 				return stats, err
 			}
--- a/TreeDB/db/vlog_gc.go
+++ b/TreeDB/db/vlog_gc.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"sync"
 
 	"github.com/snissn/gomap/TreeDB/internal/valuelog"
 )
@@ -19,6 +20,32 @@ const valueLogKeepRecentSegmentsPerLane = 2
 // second no-refresh read after the fallback Refresh call.
 var valueLogGCPostRefreshCurrentSetNoRefresh = func(vm *valuelog.Manager) *valuelog.Set {
 	return vm.CurrentSetNoRefresh()
+}
+
+var valueLogGCPostScanHook struct {
+	mu sync.Mutex
+	fn func()
+}
+
+func registerValueLogGCPostScanHook(hook func()) func() {
+	valueLogGCPostScanHook.mu.Lock()
+	prev := valueLogGCPostScanHook.fn
+	valueLogGCPostScanHook.fn = hook
+	valueLogGCPostScanHook.mu.Unlock()
+	return func() {
+		valueLogGCPostScanHook.mu.Lock()
+		valueLogGCPostScanHook.fn = prev
+		valueLogGCPostScanHook.mu.Unlock()
+	}
+}
+
+func runValueLogGCPostScanHook() {
+	valueLogGCPostScanHook.mu.Lock()
+	hook := valueLogGCPostScanHook.fn
+	valueLogGCPostScanHook.mu.Unlock()
+	if hook != nil {
+		hook()
+	}
 }
 
 // ValueLogGCOptions controls value-log garbage collection.
@@ -131,23 +158,40 @@ func (db *DB) valueLogGC(ctx context.Context, opts ValueLogGCOptions, lockMainte
 	if vm == nil {
 		return stats, fmt.Errorf("value log manager unavailable")
 	}
-	if !opts.DryRun {
-		// Publish preparation can flush value-log segments before the root that
-		// references them is installed. Serialize the reachability scan and delete
-		// phase with that prepared-publish window so GC cannot classify those
-		// segments against the previous root and remove them prematurely.
-		db.publishPrepareMu.Lock()
-		defer db.publishPrepareMu.Unlock()
-	}
-
 	observedOnly := len(opts.ObservedSourceFileIDs) > 0 && opts.ObservedSourceAssumeUnreferenced
 	var referenced map[uint32]struct{}
+	var scannedSeq uint64
 	if !observedOnly {
-		var err error
-		referenced, err = db.referencedValueLogSegments(ctx)
-		if err != nil {
-			return stats, err
+		// A full scan is O(N), so it must never run while publishPrepareMu is
+		// exclusively held. A root published after this scan can first-reference
+		// an old segment, therefore a changed commit sequence is retried outside
+		// the lock once and then conservatively aborts this GC pass.
+		const maxStaleScanRetries = 1
+		for attempt := 0; ; attempt++ {
+			var err error
+			referenced, _, scannedSeq, err = db.referencedValueLogSegmentsWithSourceAtSeq(ctx)
+			if err != nil {
+				return stats, err
+			}
+			runValueLogGCPostScanHook()
+			if opts.DryRun {
+				break
+			}
+
+			db.publishPrepareMu.Lock()
+			if db.currentCommitSeq() == scannedSeq {
+				break
+			}
+			db.publishPrepareMu.Unlock()
+			if attempt == maxStaleScanRetries {
+				return stats, nil
+			}
 		}
+	} else if !opts.DryRun {
+		db.publishPrepareMu.Lock()
+	}
+	if !opts.DryRun {
+		defer db.publishPrepareMu.Unlock()
 	}
 
 	// Prefer no-refresh snapshots to avoid repeated filesystem scans on the hot
--- a/TreeDB/db/vlog_gc.go
+++ b/TreeDB/db/vlog_gc.go
@@ -163,9 +163,11 @@ func (db *DB) valueLogGC(ctx context.Context, opts ValueLogGCOptions, lockMainte
 	var scannedSeq uint64
 	if !observedOnly {
 		// A full scan is O(N), so it must never run while publishPrepareMu is
-		// exclusively held. A root published after this scan can first-reference
-		// an old segment, therefore a changed commit sequence is retried outside
-		// the lock once and then conservatively aborts this GC pass.
+		// exclusively held. Safety invariant: GC never zombifies a segment from a
+		// referenced set older than the root visible while the exclusive lock is
+		// held. A post-scan root can first-reference an old segment, so a changed
+		// commit sequence is retried outside the lock once and then conservatively
+		// aborts this GC pass.
 		const maxStaleScanRetries = 1
 		for attempt := 0; ; attempt++ {
 			var err error
@@ -194,6 +196,9 @@ func (db *DB) valueLogGC(ctx context.Context, opts ValueLogGCOptions, lockMainte
 		defer db.publishPrepareMu.Unlock()
 	}
 
+	// Commit-sequence equality fences published roots, not prepared output. The
+	// current/recent and pending-append keep sets below are therefore computed
+	// from the value-log topology while publish preparation remains excluded.
 	// Prefer no-refresh snapshots to avoid repeated filesystem scans on the hot
 	// path. Fall back to a refresh if the manager has not yet discovered any
 	// segments (or if another process created segments on disk).
--- a/TreeDB/db/vlog_gc_incremental.go
+++ b/TreeDB/db/vlog_gc_incremental.go
@@ -237,6 +237,7 @@ func (d *valueLogRefDelta) forEachChange(fn func(fileID uint32, change int64) er
 
 type valueLogRefTracker struct {
 	mu        sync.RWMutex
+	revision  uint64
 	commitSeq uint64
 	counts    map[uint32]uint64
 	valid     bool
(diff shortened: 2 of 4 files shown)


snissn:
@codex review

snissn:
@coderabbitai review

snissn:
CI disposition for exact head `ad9469c68e`: the only failing check is `windows-core-2`, and two attempts failed in different untouched `TreeDB/internal/raftcluster` tests with the same transient `commit ambiguous / leadership lost while committing log` setup behavior. All changed-package, Linux, macOS, other Windows, perf, and both race shards passed. The Windows harness instability is now tracked in #3665 with deterministic TDD gates. One final isolated failed-job rerun is active; no code change or review reset is warranted for this unrelated failure.

PR #3686, opened 2026-07-10
perf(treedb): feed value-log ref deltas from apply

### LINKED ISSUE #3643: [PL-07] Feed value-log ref deltas from apply path instead of per-key lookups
Parent: https://github.com/snissn/gomap/issues/3630

## Goal

Fix `PL-07`: remove the extra pre-image point lookup per foreground batch entry that exists only to maintain incremental value-log ref counts.

Done means ref-delta accounting is fed by the zipper/apply path that already reads old leaf entries, or by a measured lazy strategy, so tracker-enabled writes are no longer materially slower than tracker-disabled writes for overwrite/delete workloads.

## Current Evidence

`TreeDB/db/vlog_gc_incremental.go` implements `buildValueLogRefDelta`. For each batch entry outside delete ranges it calls `lookupValueLogRefAtKey(tr, key)`, which does `tr.GetEntry(key)` against the old root. Delete ranges run range iteration. The zipper apply path will visit the same old leaves to apply the mutation, so this is duplicate foreground work.

The tracker avoids later full GC scans, so this issue must measure the tradeoff rather than blindly disabling the tracker.

## Required Fix

- Add an instrumentation milestone first: measure tracker-enabled vs tracker-disabled overwrite/delete throughput and identify the exact cost of `lookupValueLogRefAtKey`.
- Preferred implementation: teach the zipper/apply path to emit old pointer refs for replaced/deleted entries as it applies a batch, and feed `valueLogRefDelta` from that stream.
- Alternative implementation: lazy ref delta maintenance. Accumulate affected keys/ranges and resolve old refs only when GC needs a valid tracker, if profiling shows foreground accounting is not worth doing eagerly.
- Keep fail-safe behavior: tracker underflow or mismatch invalidates the tracker and forces a full scan; it must never cause unsafe deletion.
- Preserve current outer-leaf mode behavior where the tracker is disabled.

## TDD Plan

Required tests:

- [ ] Delta parity test: after mixed puts, overwrites, deletes, and delete-ranges, incremental tracker referenced set equals `scanValueLogRefCounts`.
- [ ] Old-ref emission test: replacing a pointer decrements the old file and increments the new file without `lookupValueLogRefAtKey` on the per-key path. Use a hook/counter.
- [ ] Delete-range test: range deletes decrement all old pointer refs exactly once.
- [ ] Failure/rollback test: if commit finalization fails, tracker counts roll back exactly as existing tests require.
- [ ] Existing `vlog_gc_incremental_delta_test.go`, `vlog_gc_test.go`, command-WAL publish tests, and maintenance-root tracker tests remain green.

Validation command:

```sh
GOWORK=off go test ./TreeDB/db -run 'TestValueLogGC_Incremental|TestValueLogRefTracker|TestCommandWAL|TestOrderedRoot|TestMaintenanceRoots' -count=1
GOWORK=off go test -race ./TreeDB/db -run 'TestValueLogGC_Incremental|TestValueLogRefTracker' -count=1
```

## Performance Gates

| Gate | Current | Target | Required evidence | Failure action |
| --- | ---: | ---: | --- | --- |
| per-key lookup count | one old-root `GetEntry` per batch entry | zero separate point lookups on apply-fed path | hook/counter test | block merge for preferred implementation |
| overwrite throughput | suspected tracker tax | tracker-enabled overwrite throughput within 5% of tracker-disabled, or at least 20% faster than baseline tracker-enabled | before/after benchmark | choose lazy path or keep issue open |
| allocation | duplicate lookup may allocate | no material allocation regression; target <= baseline `B/op` and `allocs/op` | benchmem | profile/tune |
| tracker correctness | full scan fallback protects safety | incremental counts equal full scan after mixed workloads | parity tests | block merge |

Required benchmark:

```sh
GOWORK=off go test ./TreeDB/db -run '^$' -bench 'BenchmarkValueLogRefDelta|BenchmarkBatchOverwrite|BenchmarkBatchDeleteRange' -benchmem -count=5
```

If benchmark names do not exist, add them. Include tracker-enabled baseline, tracker-disabled ceiling, and candidate rows with identical fixture size and batch shape.

## Non-Goals

- Do not enable the tracker for outer-leaf mode in this issue.
- Do not weaken tracker invalidation on underflow/mismatch.
- Do not claim an optimization if measurements show the tracker cost is not material; in that case, link a measured follow-up or close only with coordinator acceptance.

## PR Requirements

Use a topic branch. PR body must include instrumentation results, exact old-ref accounting design, parity tests, rollback tests, and before/after benchmarks. Latest-head CI and mature AI reviews are required before mergeability is claimed.

---

## Review corrections (Fable audit, 2026-07-09) — Verdict: MINOR

Citations verified accurate (`buildValueLogRefDelta` at `vlog_gc_incremental.go:547-591`; per-entry `lookupValueLogRefAtKey`→`tr.GetEntry` `:578`/`:603-618`; per-delete-range projection `:559-575`; underflow fail-safe `:285-303` → `invalidate()` at `db.go:2765-2768`; outer-leaf disable `:218-228`,`:551-553`). "Measure first" discipline matches the punch-list `suspected-needs-profiling` confidence.

**Corrections applied:**
1. **"Current Evidence" is imprecise:** the delta is built **after** `idx.zipper.Apply(...)` has already run (both call sites: `batch.go:388` and `batch.go:605`), against the pre-image root via COW — not "the zipper apply path *will* visit the same old leaves." The duplication claim stands, but the preferred fix requires *capturing* old-pointer info during an apply that currently emits none, not merely reordering.

2. **Preferred-path scope is underestimated:** emission must thread through all apply variants in `TreeDB/zipper` — plain `Apply`, span-native `ApplyWithOptions` (both branches at `batch.go:583-595`), and the parallel `zipper.ApplyWorkerPool` (`db.go:414`), the last requiring race-safe exactly-once aggregation across concurrent subtree workers. List these as in-scope so the implementer sizes it.

3. **Lazy alternative conflicts with the current invalidation contract:** `finalizeCommitPostWork` invalidates the tracker whenever a commit carries a nil delta (`db.go:2769-2771`). A "resolve later" design must replace that nil-delta-invalidate rule with a "pending keys" validity state — call this out.

4. **Preserve the `positives` coupling:** `valueLogRefDelta.positives` (`:155-178`) is consumed by `releasePendingValueLogAppendFileIDsFromDelta` (`value_log_appender.go:240-249`) for pending-append pin release. The refactor must preserve exact positive-count semantics (or consciously retire the path) — miscounts leak pins (GC starvation) or over-release (unsafe).

5. **Semantics to preserve:** the `:577` delete-range containment skip (avoids double-decrement when a key is both in `entries` and inside a range), and per-pointer counting for grouped/shared records.

snissn:
Closes #3643
Parent: #3630

## Problem

Tracked foreground batches rebuilt the value-log reference delta after zipper apply by calling `GetEntry` once per point mutation against the pre-image root. A 64-key overwrite test observed 64 redundant lookups, and the paired baseline benchmark showed a large tracker tax.

## Design

- Add opt-in `CollectOldPointerRefs` apply evidence with an inline-by-default count accumulator keyed by pointer file ID.
- Count removed pointer references at the existing leaf ownership point used for dead-byte accounting.
- Keep serial apply attempt-local.
- Give generic parallel child jobs and span-native work ranges private accumulators; merge them on the coordinator only after every worker succeeds.
- Build DB `valueLogRefDelta` negatives from successful apply evidence and preserve positive put counts for pending-append pin release.
- Retain the old range/point lookup builder as a fail-safe fallback when apply evidence is absent.
- Keep the tracker disabled for outer-leaf mode and preserve underflow/mismatch invalidation.

This preserves delete-range exactly-once behavior because the apply engine emits a removed entry only when it actually consumes that old leaf entry. Failed applies, optimistic retries, and failed finalization never publish the attempt-local counts.

## TDD and correctness evidence

New tests cover:

- zero foreground `lookupValueLogRefAtKey` calls (`64 -> 0`)
- serial overwrite, point-delete, and range-delete emission
- 9,000 parallel worker-pool overwrites counted exactly once
- 1,500 span-native overwrites counted exactly once
- mixed point/range tracker parity against a full scan
- grouped/shared pointer multiplicity
- inline-to-map accumulator promotion and merge
- legacy lookup fallback instrumentation
- existing failed-finalize tracker rollback

Commands:

```text
GOWORK=off GOMEMLIMIT=4GiB GOMAXPROCS=2 go test -p 1 ./TreeDB/db -run 'TestValueLogGC_Incremental|TestValueLogRefTracker|TestCommandWAL|TestOrderedRoot|TestMaintenanceRoots|TestBuildValueLogRefDeltaFallsBackWhenApplyEvidenceMissing' -count=1
PASS (latest head, 4.595s)

GOWORK=off GOMEMLIMIT=4GiB GOMAXPROCS=2 go test -race -p 1 ./TreeDB/zipper -run 'TestApplyWithOptions(CollectsOldPointerRefsForPointsAndRange|ParallelCollectsOldPointerRefsExactlyOnce|SpanNativeCollectsOldPointerRefsExactlyOnce)|TestPointerRefCountsInlinePromotionAndMerge' -count=3
PASS (latest head, 1.828s)

GOWORK=off GOMEMLIMIT=4GiB GOMAXPROCS=2 go test -race -p 1 ./TreeDB/db -run 'TestValueLogGC_Incremental|TestValueLogRefTracker|TestBuildValueLogRefDeltaFallsBackWhenApplyEvidenceMissing' -count=1
PASS

GOWORK=off GOMEMLIMIT=4GiB GOMAXPROCS=2 go test -p 1 ./TreeDB/db -count=1
PASS (354.943s)

GOWORK=off GOMEMLIMIT=4GiB GOMAXPROCS=2 go test -p 1 ./TreeDB -count=1
PASS (55.316s)

GOWORK=off go vet ./TreeDB/db ./TreeDB/zipper
PASS

GOWORK=off GOOS=windows GOARCH=amd64 go test -c ./TreeDB/db
GOWORK=off GOOS=windows GOARCH=amd64 go test -c ./TreeDB/zipper
PASS
```

The full-package runs were on the implementation commit before rebasing over the independent external-MVCC codec merge. The issue-focused and race gates were rerun on the current head.

## Performance evidence

Identical benchmark source on base `e05733d85` and candidate, `GOMAXPROCS=2`, `-benchtime=500ms -count=8`, analyzed with `benchstat`:

| row | base | candidate | delta |
| --- | ---: | ---: | ---: |
| tracker enabled | 515.9 us/op | 270.3 us/op | **-47.61%**, p=0.000 |
| tracker disabled | 195.5 us/op | 235.7 us/op | no significant change, p=0.130 |
| tracker-enabled bytes | 43.94 KiB/op | 44.01 KiB/op | +0.17% |
| tracker-enabled allocs | 99/op | 99/op | unchanged |

This clears the issue's alternate gate of at least 20% faster than tracker-enabled baseline. It does not claim the stricter within-5%-of-disabled ceiling.

Required exact regex, `-benchmem -count=5`, passed in 332.954s. Candidate medians:

| row | median |
| --- | ---: |
| tracker-enabled overwrite | 234.0 us/op, 45,058 B/op, 99 allocs/op |
| tracker-disabled overwrite | 188.8 us/op, 44,752 B/op, 98 allocs/op |
| range delete 4,096 | 376.8 us/op |
| point delete 4,096 | 675.2 us/op |

```text
GOWORK=off GOMAXPROCS=2 go test ./TreeDB/db -run '^$' -bench 'BenchmarkValueLogRefDelta|BenchmarkBatchOverwrite|BenchmarkBatchDeleteRange' -benchmem -count=5
```

### DIFF
--- a/TreeDB/db/batch.go
+++ b/TreeDB/db/batch.go
@@ -343,6 +343,7 @@ func (b *Batch) writeOptimistic(sync bool, intent *commandWALBatchIntent, maxEnt
 	tracker := newAllocTracker(idx.allocator)
 	z := idx.zipper.CloneWithAllocator(tracker)
 	applyOpts := b.flushApplyOptions()
+	applyOpts.CollectOldPointerRefs = b.db.shouldCollectValueLogRefDelta(baseSeq)
 	prepareBuf := b.db.acquireFlushApplyReadOnlyPrepareBuffer(applyOpts)
 	if prepareBuf != nil {
 		applyOpts.ReadOnlyPrepare = prepareBuf.opts
@@ -385,7 +386,7 @@ func (b *Batch) writeOptimistic(sync bool, intent *commandWALBatchIntent, maxEnt
 		hook()
 	}
 	entries, ranges := b.batch.ApplyPlan()
-	vlogRefDelta, err := b.db.buildValueLogRefDelta(idx.pager, rootID, baseSeq, entries, ranges)
+	vlogRefDelta, err := b.db.buildValueLogRefDelta(idx.pager, rootID, baseSeq, entries, ranges, &applyResult.OldPointerRefs, applyResult.OldPointerRefsCollected)
 	if err != nil {
 		b.db.releasePendingValueLogAppendFileIDsFromBatch(b.batch)
 		b.db.observeRawBatchSpanNativePublishFallback(rawSpanPlan, spanNativePublishSnapshot, FlushSpanRunFallbackOutputOwnershipFailure)
@@ -568,6 +569,7 @@ func (b *Batch) writeSerialized(sync bool, intent *commandWALBatchIntent, maxEnt
 	}
 
 	applyOpts := b.flushApplyOptions()
+	applyOpts.CollectOldPointerRefs = b.db.shouldCollectValueLogRefDelta(baseSeq)
 	prepareBuf := b.db.acquireFlushApplyReadOnlyPrepareBuffer(applyOpts)
 	if prepareBuf != nil {
 		applyOpts.ReadOnlyPrepare = prepareBuf.opts
@@ -602,7 +604,7 @@ func (b *Batch) writeSerialized(sync bool, intent *commandWALBatchIntent, maxEnt
 		return err
 	}
 	entries, ranges := b.batch.ApplyPlan()
-	vlogRefDelta, err := b.db.buildValueLogRefDelta(idx.pager, rootID, baseSeq, entries, ranges)
+	vlogRefDelta, err := b.db.buildValueLogRefDelta(idx.pager, rootID, baseSeq, entries, ranges, &applyResult.OldPointerRefs, applyResult.OldPointerRefsCollected)
 	if err != nil {
 		b.db.releasePendingValueLogAppendFileIDsFromBatch(b.batch)
 		b.db.observeRawBatchSpanNativePublishFallback(rawSpanPlan, spanNativePublishSnapshot, FlushSpanRunFallbackOutputOwnershipFailure)
--- a/TreeDB/db/flush_apply_options.go
+++ b/TreeDB/db/flush_apply_options.go
@@ -30,7 +30,7 @@ func (db *DB) flushApplyOptions() zipper.ApplyOptions {
 }
 
 func flushApplyUseOptions(opts zipper.ApplyOptions) bool {
-	return opts.ParallelApplyConcurrency > 1 || opts.SpanNativeApply || opts.PrepareReadOnly
+	return opts.ParallelApplyConcurrency > 1 || opts.SpanNativeApply || opts.PrepareReadOnly || opts.CollectOldPointerRefs
 }
 
 func (db *DB) observeFlushApplyPrepareResult(result zipper.ApplyResult, err error) {
--- a/TreeDB/db/vlog_gc_incremental.go
+++ b/TreeDB/db/vlog_gc_incremental.go
@@ -19,6 +19,7 @@ import (
 	"github.com/snissn/gomap/TreeDB/page"
 	"github.com/snissn/gomap/TreeDB/pager"
 	"github.com/snissn/gomap/TreeDB/tree"
+	"github.com/snissn/gomap/TreeDB/zipper"
 )
 
 const (
@@ -52,6 +53,11 @@ var scanValueLogRefCountsHook struct {
 	fn func()
 }
 
+var lookupValueLogRefAtKeyHook struct {
+	mu sync.Mutex
+	fn func()
+}
+
 func registerScanValueLogRefCountsHook(hook func()) func() {
 	scanValueLogRefCountsHook.mu.Lock()
 	prev := scanValueLogRefCountsHook.fn
@@ -68,6 +74,27 @@ func runScanValueLogRefCountsHook() {
 	scanValueLogRefCountsHook.mu.Lock()
 	hook := scanValueLogRefCountsHook.fn
 	scanValueLogRefCountsHook.mu.Unlock()
+	if hook != nil {
+		hook()
+	}
+}
+
+func registerLookupValueLogRefAtKeyHook(hook func()) func() {
+	lookupValueLogRefAtKeyHook.mu.Lock()
+	prev := lookupValueLogRefAtKeyHook.fn
+	lookupValueLogRefAtKeyHook.fn = hook
+	lookupValueLogRefAtKeyHook.mu.Unlock()
+	return func() {
+		lookupValueLogRefAtKeyHook.mu.Lock()
+		lookupValueLogRefAtKeyHook.fn = prev
+		lookupValueLogRefAtKeyHook.mu.Unlock()
+	}
+}
+
+func runLookupValueLogRefAtKeyHook() {
+	lookupValueLogRefAtKeyHook.mu.Lock()
+	hook := lookupValueLogRefAtKeyHook.fn
+	lookupValueLogRefAtKeyHook.mu.Unlock()
 	if hook != nil {
 		hook()
 	}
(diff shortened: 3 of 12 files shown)


snissn:
@codex review

snissn:
Exact-head validation update for `443a4df1ac450e58681c5366dad565f19c7aaa8e`:

- `GOWORK=off GOMEMLIMIT=4GiB GOMAXPROCS=2 go test -p 1 ./TreeDB/db -count=1` - PASS (`ok github.com/snissn/gomap/TreeDB/db 256.026s`)
- The worktree was clean and `origin/main` remained `e84eeb5719dabb0e5c6e6d7e525a8c624fdf3cdf` for this run.

This supersedes the PR body pre-rebase caveat for the full DB package. The latest-head CI matrix and required Codex review are still pending; no merge is being attempted until those gates complete.

snissn:
## Final merge gate - PASS

Exact head: `443a4df1ac450e58681c5366dad565f19c7aaa8e`
Validated base: `e84eeb5719dabb0e5c6e6d7e525a8c624fdf3cdf`

- Performance gate: paired `n=8` tracker-enabled overwrite median `515.9 us/op -> 270.3 us/op` (`-47.61%`, `p=0.000`); `99 allocs/op` unchanged; bytes `+0.17%`. The issue alternate target of at least 20% improvement passes.
- Correctness/performance behavior: legacy pre-image lookups `64 -> 0`; serial, generic-parallel, span-native, point-delete, range-delete, shared-pointer multiplicity, rollback, and legacy-fallback tests pass.
- Exact-head local validation: full `./TreeDB/db` passed in `256.026s`; focused race tests, full zipper, top-level TreeDB, vet, Windows compile, benchmark matrix, and diff checks passed.
- Latest-head CI: all 26 checks pass, including both TreeDB race shards, Linux/macOS/Windows suites, perf smoke, profile modes, and snapshot guardrail.
- Flake disposition: the first Root macOS attempt saw unrelated `TreeDB/caching` scheduler test `TestVlogGenerationRewriteQueue_CheckpointKickDebtDrainCapsSingleSegment` report 2 calls instead of 1. The PR changes no caching/scheduler files; the focused test passed 20/20 on this head and 20/20 on current main, the full caching package passed locally, and targeted Root macOS attempt 2 passed.
- Review: Codex completed cleanly on reviewed commit `443a4df1ac`; CodeRabbit was rate-limited and produced no findings; review-thread inventory is empty.

All issue, performance, test, CI, and review gates are satisfied. Coordinator merge authorized by the #3630 execution request.

